PR #101, opened 2026-02-21
Read SSH username from ~/.ssh/config for bip scout

### LINKED ISSUE #99: `bip scout` fails with "authentication failed" when remote SSH username differs from local OS username
`bip scout` always uses the local OS username (via `os/user.Current()`) as the SSH username. It does not read `~/.ssh/config`, so per-host `User` directives are ignored. On institutional clusters where remote usernames commonly differ from local macOS usernames, this causes all connections to fail:

```
"SSH authentication failed for server. Check ~/.ssh/config and ensure your key is authorized"
```

The error message suggests a key problem, but the actual cause is a username mismatch. The keys are fine — the server simply rejects the unknown username.

## Who is affected

Anyone whose remote SSH username differs from their local OS username. This is common in academic and institutional environments:

| Local (macOS) | Remote (cluster) |
|---------------|------------------|
| `johndoe` | `jdoe` |
| `jane.smith` | `jsmith` |

Users whose local and remote usernames match are not affected.

## Root cause

`internal/scout/ssh.go` unconditionally uses the OS username:

```go
username := ""
if u, err := user.Current(); err == nil {
    username = u.Username
}
```

The `SSHConfig` struct in `internal/scout/types.go` has no `user` field, so there is no way to override this.

## Proposed fix

Add an optional `user` field to `SSHConfig`. When set, use it for all SSH connections (both proxy jump and target servers). When omitted, fall back to the OS username so existing configurations are unaffected.

### `internal/scout/types.go`

```go
type SSHConfig struct {
    User           string `yaml:"user,omitempty"`
    ProxyJump      string `yaml:"proxy_jump,omitempty"`
    ConnectTimeout int    `yaml:"connect_timeout,omitempty"`
}
```

### `internal/scout/ssh.go`

```go
// Use configured SSH user, falling back to current OS user
username := cfg.User
if username == "" {
    if u, err := user.Current(); err == nil {
        username = u.Username
    }
}
```

### `servers.yml` usage

```yaml
servers:
  - name: server
  - pattern: "anotherserver{01..05}"
    has_gpu: true

ssh:
  user: jdoe                      # Remote SSH username
  proxy_jump: login.example.com
```

The `user` field is applied to all server connections, including the proxy jump host.

## How to diagnose

1. Check what username `bip scout` sends without `ssh.user`:
   ```bash
   whoami
   ```

2. Compare with what the remote servers expect:
   ```bash
   ssh your-server whoami
   ```
   Or check your SSH config:
   ```bash
   grep -i "user " ~/.ssh/config
   ```

3. If the two differ, this bug is the cause.

## Updated SSHConfig reference

| Field | Required | Default | Description |
|-------|----------|---------|-------------|
| `user` | no | OS username | SSH username for all connections (proxy and targets) |
| `proxy_jump` | no | (none) | Bastion/jump host for ProxyJump connections |
| `connect_timeout` | no | 10 | Timeout in seconds per connection |

matsen:
🤖

## Summary

- Parses `~/.ssh/config` to resolve per-host `User` directives, so `bip scout` respects the same username configuration as `ssh(1)`
- Resolves username at both the proxy jump level (in `NewSSHClient`) and per-target-server level (in `RunCommand`) for full fidelity
- Falls back to OS username when no config match is found, preserving existing behavior

Closes #99

## Test plan

- [x] Unit tests for SSH config parsing: exact match, wildcard, multi-pattern, first-match-wins, empty config, empty hostname
- [x] All existing tests pass (`go test ./...`)
- [ ] Manual verification on a system where remote SSH username differs from local OS username

### DIFF
--- a/internal/scout/ssh.go
+++ b/internal/scout/ssh.go
@@ -5,12 +5,62 @@ import (
 	"net"
 	"os"
 	"os/user"
+	"path/filepath"
 	"strings"
 	"time"
 
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/crypto/ssh/agent"
 )
+
+// sshConfigUser returns the SSH User directive from ~/.ssh/config for the
+// given host, or "" if none is found or the file cannot be read.
+// It matches the first Host block whose pattern matches the given hostname.
+func sshConfigUser(hostname string) string {
+	if hostname == "" {
+		return ""
+	}
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return ""
+	}
+	data, err := os.ReadFile(filepath.Join(home, ".ssh", "config"))
+	if err != nil {
+		return ""
+	}
+	return parseSSHConfigUser(string(data), hostname)
+}
+
+// parseSSHConfigUser extracts the User directive for hostname from SSH config content.
+func parseSSHConfigUser(configData, hostname string) string {
+	var inMatchingBlock bool
+	for _, line := range strings.Split(configData, "\n") {
+		line = strings.TrimSpace(line)
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
+		fields := strings.Fields(line)
+		if len(fields) < 2 {
+			continue
+		}
+		key := strings.ToLower(fields[0])
+
+		if key == "host" {
+			inMatchingBlock = false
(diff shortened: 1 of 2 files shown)


matsen:
Added!

PR #127, opened 2026-04-10
Fix concept-edge queries returning 0 results

### LINKED ISSUE #126: Concept-edge queries return 0 results due to missing concept: prefix
## Bug

`bip concept papers`, `bip paper concepts`, `bip edge list --concept`, and `bip concept delete` (edge count check) all return 0 results for concept-related edges, even when edges exist.

## Root Cause

`bip edge add` stores concept targets with a `concept:` prefix on the target ID (e.g., `concept:manifold-learning`), which is correct per the `parseNodeType`/`validateEdgeEndpoints` logic in `cmd/bip/edge.go`. However, the query functions pass bare concept IDs (e.g., `manifold-learning`), so the `WHERE target_id = ?` clauses never match.

Affected functions:
- `GetPapersByConcept` — queries with bare ID, edges store `concept:X`
- `GetConceptsByPaper` — `IN (SELECT id FROM concepts)` compares `concept:X` against bare `X`
- `CountEdgesByTarget` — called from concept delete with bare ID
- `runEdgeListByConcept` — calls `GetEdgesByTarget` with bare ID

Note: `GetEdgesByProject` already handles this correctly by prepending `project:` (line 167 of `edges_sqlite.go`).

## Repro

```bash
bip edge add -s Paper1 -t concept:my-concept -r introduces -m "test"
bip concept papers my-concept   # returns 0
bip edge list --concept my-concept  # returns 0
```

## Expected

Both commands should return the edge added above.

katosh:
Fixes #126

## Summary

- `bip concept papers`, `bip paper concepts`, `bip edge list --concept`, and `bip concept delete` all silently return 0 results for concept edges because query functions pass bare concept IDs while `edge add` stores them with a `concept:` prefix.
- Add `ensureConceptPrefix` helper and normalize IDs in `GetPapersByConcept`, `GetConceptsByPaper`, `CountEdgesByTarget`, and `runEdgeListByConcept`.
- Update test fixtures to use `concept:`-prefixed target_ids matching actual `edge add` output.

## Test plan

- [x] `TestGetPapersByConcept` — passes with prefixed edge data
- [x] `TestGetConceptsByPaper` — passes with prefixed edge data
- [x] `TestCountEdgesByTarget` — passes with prefixed edge data
- [x] Full `go test ./...` passes

### DIFF
--- a/cmd/bip/edge.go
+++ b/cmd/bip/edge.go
@@ -711,7 +711,12 @@ func runEdgeListByProject(db *storage.DB, projectID string) error {
 
 // runEdgeListByConcept outputs edges involving a specific concept.
 func runEdgeListByConcept(db *storage.DB, conceptID string) error {
-	edges, err := db.GetEdgesByTarget(conceptID)
+	// Ensure "concept:" prefix so the query matches stored edge target_ids.
+	prefixedID := conceptID
+	if !strings.HasPrefix(prefixedID, "concept:") {
+		prefixedID = "concept:" + conceptID
+	}
+	edges, err := db.GetEdgesByTarget(prefixedID)
 	if err != nil {
 		exitWithError(ExitDataError, "querying edges: %v", err)
 	}
--- a/internal/storage/concepts_sqlite.go
+++ b/internal/storage/concepts_sqlite.go
@@ -9,6 +9,14 @@ import (
 	"github.com/matsen/bipartite/internal/concept"
 )
 
+// ensureConceptPrefix returns the ID with a "concept:" prefix if not already present.
+func ensureConceptPrefix(id string) string {
+	if strings.HasPrefix(id, "concept:") {
+		return id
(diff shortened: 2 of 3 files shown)


matsen:
Thank you so much for your pull request! I'm curious, do you actually think that you will use this feature? I was thinking about stripping out these concept graphs entirely. My current workflow: I just use literature search directly. Explicitly building out this graph and curating it seemed more tedious than I thought it would be in practice.

katosh:
No idea xD I am really just trying this out rn to see if and how I could use it. It seems super cool but if it's useful, I don't know.

katosh:
Closing in favor of #128 which was merged and fixes the same issue (#126) with broader coverage (additional fixes in `edges_sqlite.go` and `GetPapersByProjectTransitive`, plus a dedicated regression test).

PR #147, opened 2026-05-15
Preserve PMCID and other resolved fields on re-import

matsen:
🤖

## Summary

`bip import --format paperpile` previously did a wholesale replace on every matched ref, silently wiping any field the Paperpile export didn't carry — notably `PMCID`, `PMID`, `ArXivID`, `S2ID`, and PDF paths added by `bip s2 linkpub`. Surfaced while building #146 (`bip ncbi backfill`): the freshly-backfilled PMCIDs would all evaporate on the next routine re-import.

Reproduced live on the real nexus library: **12 freshly-backfilled PMCIDs → 0 after one `bip import` re-run**.

## Fix

New `reference.MergeUpdate(existing, incoming) Reference`. On update, an incoming non-zero field wins (standard update behavior); an incoming zero field falls back to the existing value. `ID` and `Source` are always taken from incoming because they describe the import event itself, not the ref's content.

Wired into `persistImports` at the single update site (`cmd/bip/import.go`).

## Trade-off worth knowing

The merge preserves a field whenever the incoming side is empty. So if you clear a tag in Paperpile, the next re-import won't propagate that clearing — the incoming empty `Tags` falls back to the existing tags. Same for clearing a note, abstract, etc. This is the price of preserving externally-resolved fields. A true wipe now requires editing `refs.jsonl` directly. Given the alternative (silent loss of every backfilled PMCID), this is the right default.

## Verification

End-to-end on the real nexus, `git reset --hard` between runs:

**Before fix:**
```
PMCIDs before backfill:               4
After bip ncbi backfill --limit 20:  16   (+12)
After bip import re-run:              4   (-12, wiped)
```

**After fix:**
```
PMCIDs before backfill:               4
After bip ncbi backfill --limit 20:  16   (+12)
After bip import re-run:             16   (preserved)
```

## Test plan

- [x] `TestMergeUpdate_PreservesExternalIdentifiers` — PMCID, PMID, S2ID survive when incoming doesn't carry them.
- [x] `TestMergeUpdate_IncomingOverridesWhenNonZero` — standard update behavior when both sides have a value.
- [x] `TestMergeUpdate_IDAndSourceAlwaysFromIncoming` — ID/Source describe the import event, always from incoming.
- [x] `TestMergeUpdate_PreservesPDFAndSupplementPaths` — PDF paths added by `bip s2 linkpub` are preserved.
- [x] `TestMergeUpdate_PreservesYearOnUndatedImport` — `Published.Year: 0` in incoming doesn't wipe an existing real year.
- [x] `TestMergeUpdate_NoChangeWhenIncomingMatchesExisting` — identity merge is byte-stable.
- [x] `TestPersistImports_PreservesPostImportFields` — file-level integration: round-trip through `persistImports` and `storage.ReadAll` confirms the merge actually reaches disk.

### DIFF
--- a/cmd/bip/import.go
+++ b/cmd/bip/import.go
@@ -311,20 +311,20 @@ func classifyImport(existing []reference.Reference, newRef reference.Reference) 
 	return importAction{action: "new"}
 }
 
-// persistImports writes the import results to the refs file.
+// persistImports writes the import results to the refs file. Updates use
+// reference.MergeUpdate to preserve fields the importer didn't populate
+// (notably PMCID/PMID/ArXivID/S2ID from `bip ncbi backfill` and `bip s2`),
+// rather than wholesale-replacing the existing entry.
 func persistImports(path string, existing []reference.Reference, actions []storage.RefWithAction) error {
-	// Build new refs list
 	newRefs := make([]reference.Reference, len(existing))
 	copy(newRefs, existing)
 
-	// Apply updates first
 	for _, a := range actions {
 		if a.Action == "update" {
-			newRefs[a.ExistingIdx] = a.Ref
+			newRefs[a.ExistingIdx] = reference.MergeUpdate(newRefs[a.ExistingIdx], a.Ref)
 		}
 	}
 
-	// Append new entries
 	for _, a := range actions {
 		if a.Action == "new" {
 			newRefs = append(newRefs, a.Ref)
--- a/cmd/bip/import_test.go
+++ b/cmd/bip/import_test.go
@@ -1,9 +1,11 @@
 package main
 
 import (
+	"path/filepath"
 	"testing"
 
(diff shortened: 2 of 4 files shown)


matsen:
@psathyrella, thoughts? 

> The merge preserves a field whenever the incoming side is empty. So if you clear a tag in Paperpile, the next re-import won't propagate that clearing — the incoming empty Tags falls back to the existing tags. Same for clearing a note, abstract, etc. This is the price of preserving externally-resolved fields. A true wipe now requires editing refs.jsonl directly. Given the alternative (silent loss of every backfilled PMCID), this is the right default.

psathyrella:
Yeah that seems fine. I don't know that I've ever removed a tag from a paper. In principle it would make sense, but in practice I'm just trying to desperately categorize the fire hose without drowning, not carefully parse differences in categories.

PR #151, opened 2026-05-27
Honor env vars for GitHub and Slack tokens (BIP_GITHUB_TOKEN, BIP_SLACK_TOKEN)

### LINKED ISSUE #150: Add env-var precedence for GitHub and Slack tokens
# Add env-var precedence for GitHub and Slack tokens

## Summary

Make `GetGitHubToken()` and `GetSlackBotToken()` consult environment
variables before falling back to `~/.config/bip/config.yml`. Introduce
bip-specific `BIP_GITHUB_TOKEN` and `BIP_SLACK_TOKEN` as the recommended
names, with fallbacks to the common `GITHUB_TOKEN` / `GH_TOKEN` /
`SLACK_BOT_TOKEN` conventions. This lets users keep secrets out of
plaintext config and use secrets-manager workflows (1Password `op run`,
direnv, etc.) cleanly.

## Background

`internal/config/global.go:99-109` currently defines:

```go
func GetSlackBotToken() string {
    cfg, _ := LoadGlobalConfig()
    return cfg.SlackBotToken
}

func GetGitHubToken() string {
    cfg, _ := LoadGlobalConfig()
    return cfg.GitHubToken
}
```

Both functions read **only** from `~/.config/bip/config.yml`. Docstrings
elsewhere claim environment-variable support (e.g.
`internal/flow/slack.go:36` — "from SLACK_BOT_TOKEN environment variable
or global config"), but the underlying getter never consults the
environment. Error messages also mention "set environment variable" but
no env var name actually works for these call sites.

The practical consequence is that users must keep raw tokens on disk in
`config.yml`. Secrets-manager flows like
`op run --env-file=secrets.env -- bip ...` don't help — bip ignores the
injected env vars and reads the config file anyway.

## Proposed Approach

Modify the two getters to check environment variables first, in a
documented precedence order, before falling back to the config file.

**Key design decisions:**

- **`BIP_`-prefixed names are the recommended/documented form.** The
  unprefixed `GITHUB_TOKEN` and `GH_TOKEN` are widely set by other
  tooling (gh CLI, GitHub Actions runners, dev containers) and may have
  scopes the user didn't intend for bip. A bip-prefixed name lets users
  scope bip's access explicitly without losing compatibility.
- **Common names are honored as fallbacks** so existing setups
  (`export GITHUB_TOKEN=...` in shell profiles, gh-CLI users) keep
  working with no changes.
- **Empty env vars are treated as unset.** An exported but empty
  variable should not mask the config value.
- **Backward compatible.** With no env vars set, behavior is identical
  to today.
- **No new config schema, no new flags.** The change is internal to two
  getter functions plus their docs/tests.

### Command-Line Interface

```bash
# Recommended: bip-specific names (avoid cross-tool token bleed)
export BIP_GITHUB_TOKEN="$(op read 'op://Private/bip-github-pat/token')"
export BIP_SLACK_TOKEN="$(op read 'op://Private/bip-slack-bot/token')"
bip checkin

# Or with op run + .env file:
# secrets.env
#   BIP_GITHUB_TOKEN=op://Private/bip-github-pat/token
#   BIP_SLACK_TOKEN=op://Private/bip-slack-bot/token
op run --env-file=secrets.env -- bip checkin

# Fallback: also works (compatible with gh CLI / existing setups)
export GITHUB_TOKEN=ghp_...
export SLACK_BOT_TOKEN=xoxb-...
bip checkin
```

### Programmatic API

```go
// In bip code, nothing changes at the call site:
token := config.GetGitHubToken()
// Now returns the first non-empty value found in:
//   1. $BIP_GITHUB_TOKEN
//   2. $GITHUB_TOKEN
//   3. $GH_TOKEN
//   4. config.yml `github_token` field
```

### 1. Token getters

**File:** `internal/config/global.go` (existing)

Replace the two getters with env-aware versions:

```go
// GetGitHubToken returns the GitHub token. Precedence:
//   1. $BIP_GITHUB_TOKEN (recommended)
//   2. $GITHUB_TOKEN
//   3. $GH_TOKEN
//   4. github_token in ~/.config/bip/config.yml
// Empty env vars are treated as unset.
func GetGitHubToken() string {
    for _, name := range []string{"BIP_GITHUB_TOKEN", "GITHUB_TOKEN", "GH_TOKEN"} {
        if v := os.Getenv(name); v != "" {
            return v
        }
    }
    cfg, _ := LoadGlobalConfig()
    return cfg.GitHubToken
}

// GetSlackBotToken returns the Slack bot token. Precedence:
//   1. $BIP_SLACK_TOKEN (recommended)
//   2. $SLACK_BOT_TOKEN
//   3. slack_bot_token in ~/.config/bip/config.yml
// Empty env vars are treated as unset.
func GetSlackBotToken() string {
    for _, name := range []string{"BIP_SLACK_TOKEN", "SLACK_BOT_TOKEN"} {
        if v := os.Getenv(name); v != "" {
            return v
        }
    }
    cfg, _ := LoadGlobalConfig()
    return cfg.SlackBotToken
}
```

### 2. Error messages

Audit and update messages that mention "set environment variable" or
"add to config" so they name the variables actually accepted. Known
sites:

- `cmd/bip/project_import.go:399` — "set GITHUB_TOKEN" → name `BIP_GITHUB_TOKEN` as the recommended option.
- `cmd/bip/repo.go:168,170,571,573` — same treatment.
- `internal/flow/slack.go:40` — "SLACK_BOT_TOKEN not configured; set environment variable…" → name `BIP_SLACK_TOKEN` as the recommended option, keep `SLACK_BOT_TOKEN` as a noted fallback.
- `cmd/bip/slack_history.go:158` — config-file suggestion only; add env-var suggestion alongside it.

Suggested phrasing:

> GitHub API authentication failed. Set `BIP_GITHUB_TOKEN`
> (recommended), `GITHUB_TOKEN`, or `GH_TOKEN`, or add `github_token` to
> ~/.config/bip/config.yml.

### 3. Documentation

- `README.md` — wherever token setup is mentioned, document the env-var
  options with the recommended `BIP_*` names first.
- `docs/guides/getting-started.md` (if it covers tokens) — same.
- Mention the secrets-manager use case briefly (one-line `op run`
  example) so users know this is supported.

### Cache interaction

`LoadGlobalConfig()` memoizes the parsed config in
`globalConfigCache`. The new getters call `os.Getenv` *before*
`LoadGlobalConfig()`, so the cache is untouched on the env-var path.
`ResetGlobalConfigCache()` (used by tests) remains correct.

### Ordering rationale

`BIP_*` first lets users opt into bip-specific scoping without removing
any existing env var. If they have `GITHUB_TOKEN` exported globally for
`gh`, they can set `BIP_GITHUB_TOKEN` to a more narrowly scoped PAT and
bip will prefer it.

## Testing Strategy

Unit tests in `internal/config/global_test.go`:

- `TestGetGitHubToken_EnvVarPrecedence`
  - `BIP_GITHUB_TOKEN` set → returns it
  - Only `GITHUB_TOKEN` set → returns it
  - Only `GH_TOKEN` set → returns it
  - `BIP_GITHUB_TOKEN` + `GITHUB_TOKEN` both set → returns `BIP_GITHUB_TOKEN`
  - `GITHUB_TOKEN` + `GH_TOKEN` both set → returns `GITHUB_TOKEN`
  - All three env vars unset, config has `github_token` → returns config value
  - All env vars set to empty string → falls through to config
  - No env, no config → returns empty string

- `TestGetSlackBotToken_EnvVarPrecedence` (parallel structure)
  - `BIP_SLACK_TOKEN` set → returns it
  - Only `SLACK_BOT_TOKEN` set → returns it
  - Both set → returns `BIP_SLACK_TOKEN`
  - Empty env vars → falls through to config
  - No env, no config → empty string

Test setup must use `t.Setenv` (auto-cleanup) and call
`config.ResetGlobalConfigCache()` between subtests that mutate config.

Edge cases:

- Empty exported var (`export BIP_GITHUB_TOKEN=""`) is treated as unset.
- Whitespace-only token? Not trimmed — pass through as-is. (Consistent
  with how config-file values are handled today.)

## Documentation Updates

- [ ] `README.md` — add env-var table to the token-configuration section
- [ ] Error-message audit per "Error messages" section above
- [ ] Update docstrings on the two getters to document precedence
- [ ] `internal/flow/slack.go:36` — fix the stale "env var or global config" comment to match reality

## Dependencies

None. Uses `os.Getenv` from the standard library.

### Alternative 1: Only honor `GITHUB_TOKEN` / `SLACK_BOT_TOKEN` (no `BIP_*` prefix)

**Why not chosen:** `GITHUB_TOKEN` is commonly exported in user shells
for the `gh` CLI, often with broad scopes. A user wanting to give bip a
narrower PAT has no way to express that without unsetting their global
var. The `BIP_*` prefix solves this with a one-line addition and costs
nothing.

### Alternative 2: Config-file `${ENV_VAR}` interpolation

Allow `github_token: ${BIP_GITHUB_TOKEN}` in `config.yml`. Resolves the
plaintext-on-disk concern.

**Why not chosen:** More moving parts (string interpolation, escaping
rules, error reporting for unset vars), and it forces every user to
edit YAML to opt in. Direct env-var precedence is simpler and more
discoverable.

### Alternative 3: Use a secrets-manager SDK (1Password, Vault) directly

**Why not chosen:** Couples bip to a specific tool. Env vars are the
universal lingua franca that every secrets manager already targets via
`op run`, `vault exec`, `aws-vault exec`, etc.

## Success Criteria

- [ ] `BIP_GITHUB_TOKEN=foo bip checkin` works without `github_token` in `config.yml`
- [ ] `GITHUB_TOKEN=foo bip checkin` works as a fallback
- [ ] `op run --env-file=secrets.env -- bip checkin` works end-to-end
- [ ] Existing users with only `config.yml` see no behavioral change
- [ ] Unit tests pass for all precedence cases
- [ ] Error messages name the actual env vars accepted

## Future Work

- Apply the same env-var-first pattern to `GetS2APIKey()` and
  `GetASTAAPIKey()` for consistency (out of scope here to keep the
  diff minimal, but trivial follow-up).
- Add a `bip config doctor` subcommand that reports which sources
  (env var, config file) supplied each effective value, to aid
  debugging.

## Enhancements to CLAUDE.md

Add a short note under "Repo facts" or a new "Secrets" subsection:

> **Tokens.** `BIP_GITHUB_TOKEN` and `BIP_SLACK_TOKEN` (env) take
> precedence over `github_token` / `slack_bot_token` in
> `~/.config/bip/config.yml`. Fallbacks: `GITHUB_TOKEN`, `GH_TOKEN`,
> `SLACK_BOT_TOKEN`. Don't paste raw tokens into chats or commit them
> to the config file — prefer env vars from a secrets manager.

jaredgalloway:
## Summary

- `config.GetGitHubToken()` and `config.GetSlackBotToken()` now consult environment variables before falling back to `~/.config/bip/config.yml`, so secrets can be sourced from a manager (e.g. 1Password `op run` / `op read`) instead of living plaintext on disk.
- Precedence:
  - **GitHub:** `BIP_GITHUB_TOKEN` → `GITHUB_TOKEN` → `GH_TOKEN` → config
  - **Slack:**  `BIP_SLACK_TOKEN`  → `SLACK_BOT_TOKEN` → config
- `BIP_`-prefixed names are recommended; unprefixed names are kept as fallbacks for compatibility with the `gh` CLI and existing setups.
- Backward compatible: with no env vars set, behavior is identical to today.
- Updated error messages and docstrings to name the env vars actually accepted, plus README and `docs/guides/configuration.md`.

## Test plan

- [x] Unit tests cover all precedence cases (env-only, config-only, env-wins, empty-env-falls-through, neither-set) for both getters.
- [x] `go test ./...` — full suite passes.
- [x] `go vet ./...` and `go fmt ./...` — clean.
- [x] End-to-end smoke test (off-tree binary built from this branch):
  - Live shell (`BIP_GITHUB_TOKEN` from 1Password `op read`) → bip reads the env value.
  - Empty env vars → falls through to config.
  - `BIP_GITHUB_TOKEN=A GITHUB_TOKEN=B` → returns A.
  - `BIP_GITHUB_TOKEN="" GITHUB_TOKEN=B` → returns B.

Closes #150


matsen:
🤖 Solid implementation — precedence order, empty-env handling, and the error-message audit all match the issue spec. One blocker and a few smaller gaps from the issue's checklist:

**Blocker — `TestSlackHistoryMissingToken` fails locally.** Once `GetSlackBotToken()` honors the environment, this integration test breaks on any machine where a Slack token is reachable from the environment. The `slack` command calls `godotenv.Load()` (`cmd/bip/slack.go:30`), which reads `SLACK_BOT_TOKEN` from a `.env` in the working directory. The test runs with `cmd.Dir = nexusDir`, and the nexus has exactly such a `.env`, so the "missing token" path is never exercised:

```
--- FAIL: TestSlackHistoryMissingToken
    slack_test.go:152: expected error for missing token, got success
```

(Confirmed by running the suite against this branch. It's green in clean CI, where `getNexusDir` skips the test because no `nexus_path` is configured.) Fix — strip the token env vars and run from a clean cwd so no `.env` is loaded:

```go
env := filterEnv(os.Environ(), "XDG_CONFIG_HOME")
env = filterEnv(env, "BIP_SLACK_TOKEN")
env = filterEnv(env, "SLACK_BOT_TOKEN")
cmd := exec.Command(bp, "slack", "history", "fortnight-goals")
cmd.Dir = t.TempDir() // not nexusDir — avoids godotenv picking up the nexus .env
cmd.Env = append(env, "XDG_CONFIG_HOME="+tmpConfigDir)
```

**Smaller gaps from the issue checklist:**
- `CLAUDE.md` — the issue's "Enhancements to CLAUDE.md" asks for a short Secrets note under "Repo facts" naming the env vars.
- `bip slack --help` (`cmd/bip/slack.go`) and `docs/guides/workflow-coordination.md` still say only `SLACK_BOT_TOKEN`; worth naming `BIP_SLACK_TOKEN` (recommended) there too, for consistency with the rest of the audit.

jaredgalloway:
Thanks for the review @matsen . Pushed `4c5e629` addressing all three:

- **Blocker — `TestSlackHistoryMissingToken`:** stripped `BIP_SLACK_TOKEN` and `SLACK_BOT_TOKEN` from the child env and moved `cmd.Dir` to `t.TempDir()` so `godotenv.Load()` can't grab the nexus `.env`. Confirmed the test failed locally on `63e2730` and now passes on `4c5e629` against the same nexus.
- **`CLAUDE.md`:** added a "Secrets" bullet under Repo facts naming both precedence chains (GitHub: `BIP_GITHUB_TOKEN` → `GITHUB_TOKEN` → `GH_TOKEN` → config; Slack: `BIP_SLACK_TOKEN` → `SLACK_BOT_TOKEN` → config) and the `BIP_*`/secrets-manager recommendation.
- **`cmd/bip/slack.go` Long + `docs/guides/workflow-coordination.md`:** both now name `BIP_SLACK_TOKEN` (recommended) alongside `SLACK_BOT_TOKEN`, with the config-file fallback called out.

`go fmt`, `go vet`, and `go test ./...` all clean.

PR #155, opened 2026-05-29
scout: fix CPU measurement — instantaneous per-user (/proc sampling) + overall (top delta)

### LINKED ISSUE #154: scout: per-user CPU shows lifetime average (inflated by scout's own ssh procs); overall CPU shows since-boot
## Symptom

`bip scout --human` reports per-user CPU like `dralph(219%)` on multiple nodes simultaneously, when in fact dralph has no real CPU load on any of them. The overall `CPU` column also frequently reads near 0% on nodes that are not actually idle. The same value (~220%) appears across several nodes at once, which is a tell that the measurement is metering overhead, not real load.

Reproduce:

\`\`\`
\$ bip scout --human
Server  Status  CPU    ...  Users
orca02  online   0.1%  ...  dralph(219%) root(7%)
orca03  online   0.0%  ...  dralph(220%) root(6%)
orca04  online   0.1%  ...  dralph(219%) root(8%) mbaghban(4%)
\`\`\`

Then sshing in and running \`htop\` shows nothing.

## Root cause

Both numbers in \`internal/scout/metrics.go\` use measurements that are not "current CPU usage":

1. **Per-user CPU (Users column)** — \`ps -eo user:20,%cpu\` (line 32). The \`%cpu\` field from \`ps\` is \`(CPU time used) / (wall time since process started) * 100\`, i.e. a lifetime average. For brand-new processes (the \`sshd: dralph@notty\` that bip scout itself spawns to run the command, plus the \`systemd --user\` it triggers) the denominator is tiny, so a few ms of connection-setup CPU shows as 15–25% per process. Summed across a couple of them you get ~220%. The comment \`// runs first to avoid counting scout's own commands\` shows the author was aware of the risk, but moving the command earlier doesn't help — the sshd and systemd-user that *serve* the command are already running when ps samples.

Confirmed by \`ps\` inside one of bip scout's own ssh sessions on orca02:

\`\`\`
    PID %CPU     ELAPSED CMD
2478001 19.2       00:00 /usr/lib/systemd/systemd --user
2478070 14.2       00:00 sshd: dralph@notty
\`\`\`

00:00 elapsed, ~20% reported each.

2. **Overall CPU (CPU column)** — \`top -bn1 | grep -i \"cpu(s)\" | ...\` (line 34). top's first iteration has no delta to compute against, so it reports cumulative-since-boot stats. On a long-uptime box that's been mostly idle, that reads near 0% regardless of recent load.

## Suggested fixes (both small)

Filter out very-new processes in the user query:

\`\`\`diff
- \`ps -eo user:20,%cpu --no-headers | awk '{cpu[\$1]+=\$2} END {for (u in cpu) if (cpu[u]>1.0) printf \"%s %.1f\\n\",u,cpu[u]}' | sort -k2 -rn\`,
+ \`ps -eo user:20,%cpu,etimes --no-headers | awk '\$3 > 10 {cpu[\$1]+=\$2} END {for (u in cpu) if (cpu[u]>1.0) printf \"%s %.1f\\n\",u,cpu[u]}' | sort -k2 -rn\`,
\`\`\`

The \`etimes\` field (added in \`-o\`) gives elapsed time in seconds; \`\$3 > 10\` drops anything alive under 10 seconds, which is enough to exclude scout's own sshd/systemd-user setup. ps's \`%cpu\` is still a lifetime average — long-running processes that did heavy work hours ago and are now idle will still over-report — but at least the spurious 'always 220%' floor goes away.

Take the second iteration of top:

\`\`\`diff
- \`top -bn1 | grep -i \"cpu(s)\" | awk '{print \$2}' | cut -d'%' -f1\`,
+ \`top -bn2 -d 0.5 | grep -i \"cpu(s)\" | tail -1 | awk '{print \$2}' | cut -d'%' -f1\`,
\`\`\`

The second iteration of top has a real 0.5 s delta and reports current usage. Cost: +0.5 s per host (parallelized with \`maxConcurrent=5\`, so overall scout latency goes up by ~0.5 s only on the slowest host).

## Deeper fix (more work, much more accurate)

For per-user CPU specifically, the cleanest fix is to sample \`/proc/<pid>/stat\` twice with a 0.5 s gap and compute deltas, aggregating by \`uid\`. That gives a true instantaneous reading and is also not fooled by long-idle processes that did heavy work in the past — which the etime filter above doesn't address. But that's a real code change rather than a one-liner; the etime filter buys ~80% of the practical improvement for one extra ps column.

If/when the deeper fix lands, the \`if (cpu[u]>1.0)\` threshold should probably also apply to the post-filter aggregate (so users whose only processes were ephemeral disappear from the column rather than showing 1%).

matsen:
🤖 Closes #154.

## Confirmed both bugs

1. **Per-user CPU** used `ps %cpu`, which is a lifetime average (cputime/realtime). The short-lived sshd serving scout's own session has a tiny denominator, so it reports a large %, producing the spurious "~220%" floor seen identically across nodes.
2. **Overall CPU** used `top -bn1`, whose first iteration has no prior sample and reports cumulative-since-boot stats — near 0% on long-uptime boxes regardless of recent load. (It also only parsed top's `us` field, so it ignored system/IO time entirely — see below.)

## Fix

Both numbers now come from **one `/proc` sample**: two snapshots of `/proc/stat` and every `/proc/<pid>/stat` taken 0.5s apart. The remote command only dumps raw data; Go (`parseProcSample`) does the math:

```
echo "CLK <clk_tck> SID <session> SSHD <ppid>"
stat -c '%U %n' /proc/[0-9]*                              # pid → full username
@@SCOUT_SNAP@@
cat /proc/stat /proc/[0-9]*/stat                          # snapshot 1
@@SCOUT_SNAP@@
sleep 0.5; cat /proc/stat /proc/[0-9]*/stat               # snapshot 2
```

**Per-user CPU:** difference each process's `(utime+stime)` jiffies and aggregate by user — `delta/(CLK·dt)·100`, so `100% == one core` (same unit as `ps %cpu`). Reflects current usage; not fooled by processes that ran hot earlier and are now idle.

**Overall CPU:** from `/proc/stat`'s aggregate `cpu` line, `busy% = (Δtotal − Δidle)/Δtotal·100` (idle includes iowait). This is **total busy**, which fixes a latent under-report: the old code parsed only top's `us` (user) field, so on a syscall/IO-heavy node it read ~23% when the machine was ~67% busy. (`guest`/`guest_nice` are excluded from the total since they're already counted in `user`/`nice`.)

Computing overall CPU from the same snapshots means a single 0.5s sample window for the whole check and **no `top` dependency** (which also removes top's `~/.toprc` column-order fragility).

Two details that matter:

- **Names come from `stat -c %U`**, which is untruncated and LDAP-aware. (top's `USER` column truncates to 8 chars + `+`, merging e.g. `systemd-resolve` and `systemd-timesync` into `systemd+`.)
- **Scout's own footprint is excluded.** The dump is ~300 KB, so the sshd streaming it back uses measurable CPU that would otherwise be charged to the ssh user. That sshd is the shell's parent (`$PPID`, in a different session), so it's excluded by pid; scout's session (`SID`) covers the `cat`/`stat` pipeline.

## Validation

Run end-to-end on a live node: per-user numbers track a `top` per-user reference and the ssh user no longer appears as phantom load; overall CPU tracks top's busy figure (and is ~3x the old `us`-only reading on a system-heavy box). `parseProcSample` has unit tests for the jiffy-delta math, the `/proc/stat` busy computation, comm-with-parens parsing, session/sshd exclusion, threshold, and unknown users.

## Cost

One 0.5s sample per host for both CPU numbers (down from two sequential samples in an earlier revision). Hosts are checked in parallel, so this adds ~0.5s only on the slowest host.

## Caveat

The serving-sshd exclusion assumes `$PPID` (the command shell's parent) is the sshd itself. On a host whose login shell spawns an intermediate process before running the command (a forced-command wrapper, or `tmux`/`screen`/etc. launched from `.profile`), `$PPID` would be that wrapper instead, so the sshd's output-streaming CPU could reappear as a small phantom attributed to the ssh user. This affects only scout's self-attribution, never other users' numbers, and was not observed on the test node.

### DIFF
--- a/internal/scout/metrics.go
+++ b/internal/scout/metrics.go
@@ -2,6 +2,7 @@ package scout
 
 import (
 	"fmt"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


matsen:
🤖 @psathyrella — implemented the fix for #154 here. Both bugs confirmed; per-user CPU now does `/proc` delta sampling (parsed in Go) and overall CPU uses `top -bn2`. Would appreciate your eyes on it since you filed the issue.

internal/scout/metrics.go
```diff
@@ -28,10 +76,11 @@ const (
...
 	cmds := []string{
-		// Top CPU users — runs first to avoid counting scout's own commands
-		`ps -eo user:20,%cpu --no-headers | awk '{cpu[$1]+=$2} END {for (u in cpu) if (cpu[u]>1.0) printf "%s %.1f\n",u,cpu[u]}' | sort -k2 -rn`,
-		// CPU usage
-		`top -bn1 | grep -i "cpu(s)" | awk '{print $2}' | cut -d'%' -f1`,
+		// Top CPU users — raw /proc snapshots; parseProcUserCPU does the math.
+		topUsersCmd,
+		// CPU usage. Two iterations 0.5s apart so top reports a real delta; its
+		// first iteration has no prior sample and would report since-boot stats.
+		`top -bn2 -d 0.5 | grep -i "cpu(s)" | tail -1 | awk '{print $2}' | cut -d'%' -f1`,
```
psathyrella: Heads-up surfaced while testing this branch end-to-end: the awk extractor here picks only the `us` (user-space) CPU field. A `top -bn2 -d 0.5` line on ermine reads `%Cpu(s): 25.9 us, 39.6 sy, 0.0 ni, 34.5 id, ...`, so `awk '{print $2}'` returns `25.9` and `sy` is dropped. On kernel-heavy hosts the under-report is large.

Simultaneous cross-check on ermine, scout (built from this branch) vs. independent `top -bn2 -d 0.5` kicked off in the same shell:

|              | scout `cpu_percent` | independent `top` |
|---           |---:|---:|
| overall CPU  | 25.7% | 25.9 us + 39.6 sy = **65.5%** non-idle |

(All per-user numbers from this PR matched the same independent reference within run-to-run noise — that part of the fix works great.)

This is pre-existing — the awk extractor dates to 235f874; this PR replaced `top -bn1` with `top -bn2 -d 0.5` but didn't touch the field selection. So this PR is strictly better than main either way and absolutely landable as-is; flagging in case you'd rather fix it here than as a separate follow-up.

Quick in-place fix: pick `id` and subtract — `awk '{print 100-$8}'` (the fields are `%Cpu(s): N us, N sy, N ni, N id, ...`, so the `id` value is `$8`). The cleaner option is the "possible follow-up" already noted in the PR description: read `/proc/stat` in the same sampler and compute overall from total-vs-idle jiffies, which collapses the two 0.5s sleeps into one and drops the `top` dependency.

psathyrella:
One comment above, otherwise looks great!

matsen:
Erk, sorry, I ended up refactoring the awk into go after pinging you. Can you try pulling again?

psathyrella:
Yes look great still!

PR #163, opened 2026-05-30
Add `bip slack resolve` filter for channel-mention markup

### LINKED ISSUE #156: Add `bip slack resolve` filter to substitute `<#CXXXX>` markup with channel names
# Add `bip slack resolve` filter to substitute `<#CXXXX>` markup with channel names

🤖

## Summary

A new `bip slack resolve` subcommand that reads text on stdin, writes text on stdout, and substitutes Slack channel-mention markup (`<#CXXXXXXXX>` and `<#CXXXXXXXX|alias>`) with `#channel-name`, using `sources.yml` as the only data source. Unknown IDs pass through unchanged.

## Motivation

Slack ingests messages with raw channel-mention markup. Posts in the `fortnights` store contain bullets like `<#C03T8U5RATY>: iterate with Jared on the spike data` — unreadable to humans, and a trap for models that have to guess channel names from surrounding content.

In the `/manager.fortnight` review for the cycle ending 2026-05-18, the model labeled `C03T8U5RATY` as `#spike-data` based on Hugh's bullet phrasing. Every Jared post under that same ID since January 2026 is multidms-specific (`multidms.Data`, `multidms2`, multidms PRs #148/#208/#234, "Complete draft of multidms paper"). The channel is the private `#multidms` collaboration; the wrong label shipped into the summary before being caught.

The fix belongs in `bip slack` — that's the layer that already understands Slack and reads `sources.yml`. A stdin/stdout filter is the most composable shape: every Slack-touching skill (`/manager.fortnight`, `/bip.digest`, `/bip.narrative`) becomes a one-line pipe addition.

## Configuration

The new `project_channels` block is already in production in `~/re/nexus/sources.yml`:

```yaml
slack:
  channels:
    fortnight-goals:
      id: C02GYV1GH2M
      purpose: goals
    # …additional named channels…
  project_channels:
    C08JB3LRDU2: flu-mut-rates
    C03T8U5RATY: multidms
    # …additional id→name entries…
```

Two maps feed the resolver:

- `slack.channels` (existing): keyed `name → {id, purpose}`. Reverse to `id → name`.
- `slack.project_channels` (new): keyed `id → name` directly.

Collisions across the two maps are not expected (different intents). If an ID appears in both, prefer `project_channels` and pin the behavior with a test.

## Implementation

1. **`internal/flow/types.go`** — add `ProjectChannels map[string]string \`yaml:"project_channels"\`` to `SlackConfig`.
2. **`internal/flow/slack.go`** — add `LoadChannelIDMap(nexusPath) (map[string]string, error)`. Call `LoadSources` directly (not `LoadSlackChannels`), so a `sources.yml` with only `project_channels` and no `channels` block still resolves. Return an error only if **both** maps are empty.
3. **`internal/flow/slack.go`** — add a pure helper `ResolveChannelMentions(text string, idToName map[string]string) string`. Regex `<#(C[A-Z0-9]+)(?:\|([^>]*))?>`. Replacement priority: map name → non-empty alias from markup → leave raw `<#CXXXX...>` unchanged. Pure function so it's unit-testable without Cobra, and it sits alongside the existing `ResolveUserName` precedent.
4. **`cmd/bip/slack_resolve.go`** (new, ~40 LOC) — Cobra subcommand. Loads the merged map once via `LoadChannelIDMap`, reads all of stdin with `io.ReadAll` (Slack mention markup never spans lines and fortnight-scale output is well under a MB; do NOT reach for `bufio.Scanner`, whose default 64 KB token size silently truncates), runs `ResolveChannelMentions`, writes to `os.Stdout`.
5. **`cmd/bip/slack.go`** — register the new subcommand. Extend the parent `bip slack` `--help` long description to mention `resolve`.

No skill catalog currently lists `bip slack` subcommands, so no skill-doc update is needed in this PR. The follow-up issue to teach `/manager.fortnight` / `/bip.digest` / `/bip.narrative` to pipe through `resolve` is separate.

## Test plan

Pure-function and end-to-end cases, both bug-catching:

- Known ID resolves: `<#C08JB3LRDU2>` → `#flu-mut-rates`.
- Alias form prefers map: `<#C08JB3LRDU2|whatever>` → `#flu-mut-rates`, not `#whatever`.
- Alias-only fallback: `<#CZZZZ99|fallback-name>` → `#fallback-name`.
- Empty alias on known ID: `<#C044B4JUE5U|>` → `#antigen` (alias is empty string, not "absent"; map name wins). This form actually appears in `fortnights.jsonl`.
- Empty alias on unknown ID: `<#CZZZZ99|>` → `<#CZZZZ99|>` (no usable name; pass through rather than emit a bare `#`).
- Override semantics: an ID present in both `channels` and `project_channels` resolves to the `project_channels` value.
- Adjacent markup, no whitespace: `<#C08JB3LRDU2><#C03T8U5RATY>` → `#flu-mut-rates#multidms` (catches greedy-regex bugs).
- Markup at end of input with no trailing newline: `prefix <#C03T8U5RATY>` → `prefix #multidms`.
- Markup inside a backtick code fence resolves: `` `<#C03T8U5RATY>` `` → `` `#multidms` `` — the filter is intentionally dumb about Markdown structure; callers that want literal channel IDs inside code fences should not pipe through `resolve`.
- Malformed lowercase ID passes through unchanged: `<#c03t8u5raty>` → `<#c03t8u5raty>` (Slack IDs are uppercase; regex must anchor).
- Trailing punctuation: `see <#C03T8U5RATY>.` → `see #multidms.` (no eating of the period).
- Multiple IDs in one line all resolve.
- Empty stdin → empty stdout, exit 0.

Test locations follow the repo's existing pattern:

- Pure-function cases (regex behavior, override semantics, empty-alias) → `internal/flow/slack_test.go`, named `TestResolveChannelMentions_*` and `TestLoadChannelIDMap_*`.
- End-to-end stdin/stdout → `tests/integration/slack_test.go`, named `TestSlackResolve_*`. No Slack token required — this command never touches the API.

## Success criteria

1. `echo '• <#C03T8U5RATY>: iterate on spike data' | bip slack resolve` outputs `• #multidms: iterate on spike data`. Whether `/manager.fortnight` adopts the filter is tracked in a follow-up issue.
2. `bip slack history fortnight-feats --since 2026-05-13 | bip slack resolve` replaces every `<#C...>` mention configured in `sources.yml` with `#name`, and leaves unknown IDs raw and visible.
3. New unit + integration tests pass; existing slack tests unchanged.

## Scope

In scope:

- stdin/stdout filter
- `sources.yml` as the only data source
- `bip slack` `--help` long description updated to mention `resolve`

Out of scope (separate issues if wanted):

- Live `conversations.info` lookups
- Auto-populating `project_channels` from the Slack API
- Changes to `bip store query` output formatting
- Other Slack markup (`<@U…>` user pings, `<!subteam…>`)
- Teaching `/manager.fortnight` / `/bip.digest` / `/bip.narrative` to pipe through the filter

## References

- Production `sources.yml` block: `~/re/nexus/sources.yml` (shipped 2026-05-28).
- Misattribution incident: `~/writing/manager/fortnight-summaries/2026-05-18.md` (initial draft labeled `C03T8U5RATY` as `#spike-data`; corrected to `#multidms` after evidence review).
- Existing slack code (precedent for placement and naming): `cmd/bip/slack*.go`, `internal/flow/slack.go` (`LoadSlackChannels` at line 467, `ResolveUserName` at line 501), `internal/flow/types.go` (`SlackConfig`, `SlackChannelConfig`).

matsen:
🤖

## Summary

- New `bip slack resolve` subcommand: a token-free stdin/stdout filter that rewrites Slack channel-mention markup (`<#CXXXX>` and `<#CXXXX|alias>`) to `#channel-name`, using `sources.yml` as the only data source. Unknown IDs with no usable alias pass through unchanged.
- `internal/flow/types.go`: added `ProjectChannels map[string]string` (`project_channels`) to `SlackConfig`.
- `internal/flow/slack.go`: added `LoadChannelIDMap` (merges reversed `slack.channels` with `slack.project_channels`; the latter wins on ID collision; resolves even when only `project_channels` is present) and the pure helper `ResolveChannelMentions`. Resolution priority per mention: mapped name → non-empty alias from markup → original markup unchanged.
- `cmd/bip/slack_resolve.go`: new ~60-LOC Cobra command. Reads all of stdin with `io.ReadAll` (not `bufio.Scanner`, whose 64 KB token limit would truncate). `cmd/bip/slack.go`: registered the subcommand and extended the parent `--help` long description.

The regex `<#(C[A-Z0-9]+)(?:\|([^>]*))?>` anchors to uppercase IDs, so malformed lowercase markup passes through.

## Test plan

- `internal/flow/slack_test.go`: `TestResolveChannelMentions` (13 cases: known ID, alias-prefers-map, alias-only fallback, empty-alias on known/unknown ID, adjacent markup, backticks, lowercase passthrough, trailing punctuation, …) and `TestLoadChannelIDMap_*` (merge, override, project-channels-only, empty-error).
- `tests/integration/slack_test.go`: `TestSlackResolve_*` drive the real binary end-to-end through a hermetic temp nexus — no Slack token required, so these always run (no skip).
- `go fmt`, `go vet`, and the full `go test ./...` all pass.

Smoke check against the production nexus:

```
$ echo '• <#C03T8U5RATY>: iterate on spike data' | bip slack resolve
• #multidms: iterate on spike data
```

## Out of scope (per issue)

No skill catalog currently lists `bip slack` subcommands, so no skill-doc update is needed here. Teaching `/manager.fortnight`, `/bip.digest`, and `/bip.narrative` to pipe through `resolve` is tracked separately.

Closes #156

### DIFF
--- a/cmd/bip/slack.go
+++ b/cmd/bip/slack.go
@@ -22,6 +22,10 @@ Requires a Slack bot token with channels:history, channels:read, and
 users:read scopes. Sourced from BIP_SLACK_TOKEN (recommended) or
 SLACK_BOT_TOKEN, falling back to slack_bot_token in ~/.config/bip/config.yml.
 
+The 'resolve' subcommand is a token-free stdin/stdout filter that rewrites
+channel-mention markup (<#CXXXX>) to #channel-name using sources.yml; pipe
+any Slack-derived text through it for human-readable channel references.
+
 All commands output JSON by default for agent consumption.
 Use --human flag for human-readable output.`,
 }
--- a/cmd/bip/slack_resolve.go
+++ b/cmd/bip/slack_resolve.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/matsen/bipartite/internal/config"
+	"github.com/matsen/bipartite/internal/flow"
+	"github.com/spf13/cobra"
+)
+
+var slackResolveCmd = &cobra.Command{
+	Use:   "resolve",
+	Short: "Substitute Slack channel-mention markup with channel names",
+	Long: `Read text on stdin, write text on stdout, replacing Slack channel-mention
+markup (<#CXXXXXXXX> and <#CXXXXXXXX|alias>) with #channel-name.
+
+Names come from sources.yml: the slack.channels block (reversed id->name) and
+the slack.project_channels block (id->name). When an ID appears in both,
+project_channels wins. Unknown IDs with no usable alias pass through unchanged.
+
+This is a dumb text filter that never touches the Slack API. It makes no
+assumptions about Markdown structure, so mentions inside code fences are
+resolved too.
+
+Examples:
+  echo '• <#C03T8U5RATY>: notes' | bip slack resolve
+  bip slack history fortnight-feats --since 2026-05-13 | bip slack resolve`,
+	Args: cobra.NoArgs,
+	RunE: runSlackResolve,
+}
+
+func init() {
+	slackCmd.AddCommand(slackResolveCmd)
+}
+
+func runSlackResolve(cmd *cobra.Command, args []string) error {
+	nexusPath := config.MustGetNexusPath()
+	idToName, err := flow.LoadChannelIDMap(nexusPath)
+	if err != nil {
+		return outputSlackError(1, "config_error", err.Error())
+	}
+
+	// Read all of stdin. Mention markup never spans lines and fortnight-scale
+	// input is well under a MB, so io.ReadAll is correct here; bufio.Scanner's
+	// default 64 KB token limit would silently truncate larger input.
+	input, err := io.ReadAll(os.Stdin)
+	if err != nil {
+		return outputSlackError(1, "io_error", err.Error())
+	}
+
+	resolved := flow.ResolveChannelMentions(string(input), idToName)
+	if _, err := fmt.Fprint(os.Stdout, resolved); err != nil {
+		return outputSlackError(1, "io_error", err.Error())
+	}
+	return nil
+}
--- a/internal/flow/slack.go
+++ b/internal/flow/slack.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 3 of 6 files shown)


matsen:
🤖 **Suggestion: prefer hermetic fixtures for API-free slack tests**

A pattern question worth deciding as a team, prompted by the `resolve` tests here.

Most of the existing `slack` integration tests use `setupTempDirWithSlackConfig`, which copies the real nexus `sources.yml` and `t.Skip(...)`s when no Slack config is present — appropriate for commands that hit the Slack API. But `bip slack resolve` never touches the API and takes `sources.yml` as its only input.

For that class of command, I'd suggest we standardize on a **hermetic fixture** (temp nexus + temp `XDG_CONFIG_HOME` with a controlled `project_channels` block, as in `setupResolveEnv`) rather than the copy-real-config-and-maybe-skip path. The upside:

- Tests always run — in CI and on a fresh clone — with no silent skips.
- We can pin exact override/collision semantics that the real config can't guarantee.

The cost is a little helper duplication. If folks agree this is the right default for API-free commands, a good follow-up would be to factor `setupResolveEnv` into a shared helper and migrate any other non-API slack tests onto it. Flagging it here rather than expanding scope in this PR.

PR #185, opened 2026-08-25
Add Volume/Issue/Pages fields and parse PMID from Paperpile imports

matsen:
🤖

## Summary

`bip export --bibtex` couldn't produce submission-ready BibTeX because it never emitted `volume`, `number` (issue), `pages`, or `pmid`, even for journal articles where Paperpile has all four. This adds `Volume`/`Issue`/`Pages` to `Reference`, parses them (plus `pmid`) from Paperpile imports, and emits all four in BibTeX output.

- `internal/reference/reference.go`: add `Volume`, `Issue`, `Pages` string fields
- `internal/reference/merge.go`: fallback-merge the three new fields (same pattern as `PMID`/`PMCID`)
- `internal/storage/sqlite.go`: add `volume`/`issue`/`pages` columns, thread through insert/scan
- `internal/importer/paperpile.go`: parse `volume`/`issue`/`pages`/`pmid` (via `FlexibleString`) from `PaperpileEntry`
- `internal/export/bibtex.go`: emit `volume`/`number`/`pages`/`pmid` when non-empty
- `cmd/bip/get.go`: print `Volume`/`Issue`/`Pages`/`PMID` in `--human` output

## Test plan

- `internal/importer/paperpile_test.go`: `TestParsePaperpile_VolumeIssuePagesPMID` — all four fields land on the resulting `Reference`
- `internal/export/bibtex_test.go`: `TestToBibTeX_VolumeIssuePagesPMID` (fields present) and updated `TestToBibTeX_OptionalFields` (regression check — no empty lines emitted)
- `internal/reference/merge_test.go`: `TestMergeUpdate_PreservesVolumeIssuePages` — existing values survive a re-import that omits them
- `go build ./... && go vet ./... && go test ./...` all pass

DEFERRED: none — out-of-scope items (backfilling these fields for existing references, parsing volume/issue/pages from S2/ASTA imports) are called out explicitly in the issue as follow-up work, not deferred from this PR's scope.

### DIFF
--- a/cmd/bip/get.go
+++ b/cmd/bip/get.go
@@ -65,6 +65,22 @@ func printRefDetail(ref reference.Reference) {
 		fmt.Printf("Venue:    %s\n", ref.Venue)
 	}
 
+	if ref.Volume != "" {
+		fmt.Printf("Volume:   %s\n", ref.Volume)
+	}
+
+	if ref.Issue != "" {
+		fmt.Printf("Issue:    %s\n", ref.Issue)
+	}
+
+	if ref.Pages != "" {
+		fmt.Printf("Pages:    %s\n", ref.Pages)
+	}
+
+	if ref.PMID != "" {
+		fmt.Printf("PMID:     %s\n", ref.PMID)
+	}
+
 	// Date
 	date := fmt.Sprintf("%d", ref.Published.Year)
 	if ref.Published.Month > 0 {
--- a/internal/export/bibtex.go
+++ b/internal/export/bibtex.go
@@ -45,6 +45,26 @@ func ToBibTeX(ref reference.Reference) string {
 		b.WriteString(fmt.Sprintf("  doi = {%s},\n", ref.DOI))
 	}
 
+	// Volume (optional)
+	if ref.Volume != "" {
+		b.WriteString(fmt.Sprintf("  volume = {%s},\n", ref.Volume))
+	}
+
+	// Issue/number (optional)
+	if ref.Issue != "" {
+		b.WriteString(fmt.Sprintf("  number = {%s},\n", ref.Issue))
+	}
+
+	// Pages (optional)
+	if ref.Pages != "" {
+		b.WriteString(fmt.Sprintf("  pages = {%s},\n", ref.Pages))
+	}
+
+	// PMID (optional)
+	if ref.PMID != "" {
+		b.WriteString(fmt.Sprintf("  pmid = {%s},\n", ref.PMID))
+	}
+
 	// Abstract (optional, if present)
 	if ref.Abstract != "" {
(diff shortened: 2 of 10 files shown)


matsen:
🤖 Tested this branch end-to-end against a real Paperpile export (`bip export --bibtex DeWitt2026-cw` now correctly emits `volume = {189}`, `number = {16}`, `pages = {4980-4996.e8}`, `pmid = {42248140}` — matches CrossRef). `go test ./internal/reference/... ./internal/importer/... ./internal/export/... ./cmd/bip/...` all pass.

Found one real bug while testing, unrelated to the new fields themselves: **`bip rebuild` doesn't handle schema changes** on an existing cache file.

Repro:
1. On `main` (pre-PR), run `bip rebuild` — creates `.bipartite/cache/refs.db` with the old `refs` schema (no `volume`/`issue`/`pages` columns).
2. Switch to this branch, `bip rebuild` again against the *same* cache file.
3. Fails: `SQL logic error: table refs has no column named volume (1)`.

Root cause: `createSchema` (`internal/storage/sqlite.go:84`) uses `CREATE TABLE IF NOT EXISTS refs (...)`, so once the table exists with the old columns, the new schema in code is never applied. `RebuildFromJSONL` (`internal/storage/sqlite.go:138`) only does `DELETE FROM refs` / `DELETE FROM refs_fts` — it clears rows, not schema — then the `INSERT` naming the new `volume, issue, pages` columns fails against the stale table.

This contradicts the "SQLite is ephemeral, rebuilt from JSONL, no migration needed" design principle (DESIGN.md) — in practice a schema change silently breaks `bip rebuild` for anyone with a pre-existing cache, with no guidance to delete it. Workaround I used: `rm ~/re/nexus/.bipartite/cache/refs.db` before rebuilding.

Suggest (separate from this PR, or as a small addition to it): have `OpenDB`/`createSchema` or `RebuildFromJSONL` drop and recreate the `refs`/`refs_fts` tables (`DROP TABLE IF EXISTS` + `CREATE TABLE`) rather than relying on `CREATE TABLE IF NOT EXISTS`, so future schema changes don't require users to manually delete the cache file. Happy to file this as a follow-up issue instead if you'd rather keep this PR scoped to volume/issue/pages/pmid.

matsen:
🤖 Fixed in this PR (81f7c30): `RebuildFromJSONL` now `DROP TABLE IF EXISTS` + recreates `refs`/`refs_fts` instead of `DELETE FROM`, so a rebuild always picks up the current code's schema regardless of what an older binary left behind. Added `TestDB_RebuildFromJSONL_SchemaDrift`, which creates a pre-PR-shaped `refs` table (no `volume`/`issue`/`pages` columns) and confirms rebuild against it now succeeds and returns the new fields correctly. Confirmed your repro no longer reproduces.

PR #191, opened 2026-08-27
Fix SendMessage nudge rules: durable record on worklog, no bare pointers

### LINKED ISSUE #190: Correct the SendMessage nudge rules: the dual-write mandate is silently lossy, and pointer-only messages carry no signal
# Correct the SendMessage nudge rules: the dual-write mandate is silently lossy, and pointer-only messages carry no signal

🤖

Two of the five rules `c19f9e2` added to `skills/bip-epic/SKILL.md` and `skills/bip-epic-poll/SKILL.md` are replaced with rules that hold under test; the other three stay.

## Motivation

`c19f9e2` (PR #189, closing #188) documented `ListAgents`/`SendMessage` as a nudge channel for correcting live EPIC workers. Applying it the same morning in a 12-worker phyz fleet, two of its rules failed — one of them in a way that makes the mitigation riskier than the risk it guards against. The rules are new enough that nothing depends on them yet, so this is cheap to fix now.

`CONSTITUTION.md` II requires that all state that matters survive compaction and session boundaries, with the corollary that state you cannot reconstruct from files is a bug. `c19f9e2`'s dual-write mandate is an attempt to satisfy II. This correction does not weaken II — it moves the durable record off the file the worker continuously rewrites and onto one it only appends to. The dividing line is the deliverable: any nudge that changes what the worker will produce (scope, target, artifact, gate criterion) is state that matters and MUST be recorded durably in the same step as the message. Only facts that leave the deliverable unchanged — which host to use, a peer's timing, a dependency that just landed — may be message-only, and those are precisely the facts II does not cover.

## Evidence

Measured with a disposable test worker (a clone told to full-rewrite `.epic-status.json` once per step across ten steps, no git or build operations), plus two incidental observations from real workers in the same fleet.

**The dual-write is lossy unless the worker was told to preserve the field.** Mid-flight, at the worker's step 7, the conductor injected two things in one read-modify-write: `lead_guidance` (which *this* worker's prompt explicitly told it to read and preserve) and a `conductor_probe` key (which it was never told about). After the worker's next rewrites:

- `lead_guidance` — preserved, 218 chars, and followed: the worker skipped the two directories the guidance told it to skip and substituted the two files it asked for.
- `conductor_probe` — `null`. The worker rebuilt the object from its prompt's template and dropped the key silently.

**No spawn prompt contains a preserve instruction.** The canonical status-field block that every EPIC spawn prompt is composed from (`skills/bip-epic-spawn/SKILL.md:158-174`) describes `lead_guidance` as "what the lead told you to do next (set by lead)" and lists the field set a worker maintains; nothing anywhere tells a worker to preserve fields outside that list when it rewrites the file. The recovery instructions immediately below (lines 179-184) reinforce this — they tell the worker to *read* `lead_guidance`, never to round-trip unknown keys. The `conductor_probe` arm is therefore the production configuration, and the `lead_guidance` arm only passed because the test prompt was written to make it pass.

**The pointer-only message was a measured no-op.** The worker's own report: "got your re-read ping mid-way — re-read the status file at that point, content was unchanged from what I'd already captured." The behavior change came from the file write; the ping added nothing, because the status protocol already has workers read that file at every step.

**A real worker captured a received message durably, with no dual-write.** An unrelated live worker recorded, in `.epic-worklog.md`, in the same turn the message arrived: "Cross-session message from conductor `phyz-1a` arrived asking me to adopt the issue-lead check-in protocol at stopping points … acknowledged and adopting it". Durable capture happened for free, into the append-only file rather than the contended one. Note this was voluntary, not instructed — which is why the corrected rule assigns an owner rather than relying on it.

**The address is not the tmux window name.** `SendMessage` to `test-maple`, the window name the conductor chose at spawn, returned `No agent named 'test-maple' is reachable.` The address was `maple-de` — a session name the conductor neither chooses nor can predict, discoverable only via `ListAgents`.

**`notify_when_idle` works.** A subscription on the test worker delivered one notice at 06:19 carrying the harness summary "All 10 steps are complete", with no polling.

## Problem

**Rule**: "Any correction sent via `SendMessage` MUST also be written to `lead_guidance` in the same step…" Three defects, in descending order of severity:

- It is silently lossy in the configuration the fleet actually runs, per the `conductor_probe` arm above.
- It mandates an unsynchronized read-modify-write on a file the worker owns and rewrites continuously. In the live fleet this meant a conductor merging into a 4,000-character `lead_guidance` that a worker's own issue-lead had written twenty minutes earlier — either writer can clobber the other, and the field being clobbered is the canonical instruction.
- It creates a second writer to a field every spawn prompt attributes to the lead, so a conductor nudge becomes indistinguishable from a lead verdict. When the two disagree, "`lead_guidance` remains canonical" gives no tiebreak, because both are now inside it.

**Rule**: "Send a pointer (…), not the instruction payload…" The rule appears in two wordings — `bip-epic-poll:135` "not the instruction payload" and `bip-epic:307` "rather than the instruction itself" — so an implementer must grep for both.

- It depends on the dual-write having landed intact, so it inherits the defect above: a pointer to a clobbered field is worse than a self-contained message.
- It is a no-op for a protocol-following worker, as measured.
- It strips the priority signal. "Re-read `lead_guidance`" tells the worker nothing about whether to drop what it is doing or finish the current step.
- It treats all messages as durable scope changes. Most conductor nudges are transient coordination facts — this host is loaded, the blocking PR just landed, a peer is mid-rebase on your file — whose value expires in minutes. Filing those in `lead_guidance` pollutes the field whose worth comes from being the current canonical instruction.

## Proposed implementation

1. **Replace the dual-write mandate.** Split into three prescriptions:
   - Send the correction directly — that is the channel's purpose. No status-file write is a precondition for messaging.
   - When the nudge changes the deliverable, the **conductor** appends a timestamped, attributed entry to `.epic-worklog.md` in the same step. Workers are already told to treat that file as append-only and never to edit previous entries (`skills/bip-epic-spawn/SKILL.md:175`), so a conductor append cannot destroy worker history the way a `lead_guidance` merge can.
   - If the status file is written at all, use a field the lead does not own (`conductor_guidance`, or a `lead_notes` entry tagged `source: conductor`). Never merge into `lead_guidance`.
2. **Replace the pointer rule.** Send the correction, briefly: one line of what changed is the floor, since a bare pointer is a no-op for a worker already reading the file each step. If a correction is long enough that pointing at a file seems preferable, its home is the spawn prompt or the issue body, not a nudge — length is an argument for brevity, not for indirection.
3. **Keep the other three rules**: next-tool-call delivery and "not a substitute for `tmux capture-pane`"; addressable-sessions-only with `ListAgents` first and a file-only fallback; no permission laundering, including instructions phrased as a message to a worker.
4. **Add**: the address is the session name from `ListAgents`, not the tmux window name the conductor assigned, quoting the failure text.
5. **Add**: when *not* to nudge — a worker in `awaiting-results` with a live `check_cmd` needs no ping — and `notify_when_idle: true` as the anti-polling alternative to "tell me when this worker finishes". Record its bounds: it works only from the main conversation, only for sessions on this machine, and it is one-shot (omit `message` for a pure subscription that costs the target nothing; a subscription that never fires reports as expired).

## Files to modify

- `skills/bip-epic/SKILL.md` — two sites: the `### Correcting a live worker: SendMessage as a nudge channel` section (line 46), and the Step 6 paragraph (line 300) that restates the dual-write ("update `.epic-status.json`'s `lead_guidance` … as usual") and the pointer rule ("`SendMessage` a short pointer … rather than the instruction itself"). Both must change; the Step 6 paragraph should shrink to a pointer at the section above.
- `skills/bip-epic-poll/SKILL.md` — the `**Correcting a live worker mid-flight**:` paragraph (line 124). It is not a cross-reference: it independently restates all five rules and contains no pointer to `bip-epic`, so it needs the full edit, not a trim.

## Test plan

- **Verbatim-preservation gate**, which makes criterion 5 mechanical. Extract the three kept bullets from the commit and require exact substring matches after editing, then confirm the replaced wordings are gone:

```
git show c19f9e2:skills/bip-epic/SKILL.md | sed -n '/^- Delivery is not instant/,/^$/p' > /tmp/k2.txt
git show c19f9e2:skills/bip-epic/SKILL.md | sed -n '/^- `SendMessage` only reaches/,/^$/p' > /tmp/k3.txt
git show c19f9e2:skills/bip-epic/SKILL.md | sed -n '/^- Do not use `SendMessage`/,/^$/p' > /tmp/k4.txt
for f in /tmp/k2.txt /tmp/k3.txt /tmp/k4.txt; do grep -qF -f "$f" skills/bip-epic/SKILL.md && echo "OK $f" || echo "DRIFT $f"; done
grep -nE "MUST also be written to .lead_guidance|not the instruction payload|rather than the instruction itself" skills/bip-epic/SKILL.md skills/bip-epic-poll/SKILL.md
```

  Expected: `OK` three times, zero hits from the second grep.

- **Round-trip reconstruction test** — the one check that separates "correct" from "better worded", and a direct probe of the Constitution II risk. Send a live worker a message-only correction that changes an artifact ("drop `--foo` from the benchmark you're about to run"), then spawn a **fresh** `issue-lead` (`agents/issue-lead.md`) against that clone with no inherited context — its documented inputs are `.epic-status.json` and `.epic-worklog.md` (`agents/issue-lead.md:23-24`). Ask it to state the current instruction. Expected: the lead names the correction and does not reverse it. Failure mode to watch: a deliverable-changing nudge misclassified as transient is invisible to the lead, which re-issues the original instruction — a live reproduction of the II corollary.

- **Adversarial concurrent-append test** on the substituted channel, to measure rather than assert that appending to `.epic-worklog.md` avoids the clobber `.epic-status.json` suffers. With a live worker writing at its normal cadence, have the conductor append a uniquely numbered marker 10 times at random offsets. Expected: 10/10 markers present and zero worker entries lost. If any are lost, the proposal trades a measured clobber for an unmeasured one and the guidance needs a hedge.

- **Pre-registered control**: inject a conductor-owned key into a worker's `.epic-status.json` mid-flight and confirm it is absent after the worker's next 3 status rewrites ("the next rewrite" is not deterministic). This reproduces the regression the guidance must stop recommending.

## Success criteria

1. Neither skill mandates a write to `.epic-status.json` as a precondition for sending a message.
2. Both name `.epic-worklog.md` as the durable channel for a scope correction, and require attribution distinct from `lead_guidance` if the status file is written at all.
3. Both state that the address is the `ListAgents` session name, not the tmux window name, and quote the failure text.
4. Both state when not to nudge, and name `notify_when_idle` with its three bounds.
5. The three kept rules survive as rules. In `skills/bip-epic/SKILL.md` the three bullets are byte-identical to `git show c19f9e2:skills/bip-epic/SKILL.md`, verified by exact substring match. In `skills/bip-epic-poll/SKILL.md` the same three rules are preserved in substance, with one licensed edit: the fallback clause's "edit `lead_guidance`" becomes the non-lead-owned field required by criterion 2.
6. Both skills state the durable/transient test in terms of the deliverable, and both require a durable record for any nudge that changes what the worker will produce.
7. Neither skill recommends a bare pointer message. Both require the message to state what changed in at least one line, and both say that a correction too long to send is a spawn-prompt or issue-body change, not a nudge.

## Scope boundaries

**In scope**: the `SendMessage` text in the two skills named above, at all three sites.

**Out of scope**:

- Changing the EPIC status protocol's **field set** in spawn prompts — that text lives in `bip-epic-spawn` and would need its own issue. (A one-line instruction telling workers to log received cross-session messages is in scope if the conductor-append route is rejected.)
- Adding a locking or merge protocol to `.epic-status.json`. The corrected guidance routes around the contention instead, which is cheaper and needs no worker-side change.
- The phyz-side defects found in the same session (a `remote-sync` exclude that misses the repo's own per-clone cache-dir convention, and spawn prompts that omit the issue-lead protocol). Both belong to phyz.

## References

- `c19f9e2` / PR #189 / issue #188 — the guidance this revises.
- `CONSTITUTION.md` II — the file-based-state requirement the dual-write mandate was implementing.
- `skills/bip-epic-spawn/SKILL.md:158-175` — the canonical status-field block and the worker's append-only worklog constraint.
- `PROSE-DISCIPLINE.md` — applied to this issue body.

matsen:
🤖

## Summary
`c19f9e2` (#189) documented `SendMessage` as a nudge channel for correcting live EPIC workers. Two of its five rules failed under test the same day: the dual-write mandate into `.epic-status.json`'s `lead_guidance` is silently lossy (a worker rewriting the file from its prompt template drops any key not in that template, verified with a disposable test worker), and it creates an unsynchronized second writer to a field every spawn prompt attributes to the lead. The pointer-only message rule was a measured no-op, since a protocol-following worker already re-reads the status file every step.

- Replace the dual-write mandate: send the correction directly, and durably record only nudges that change the deliverable (scope, target, artifact, gate criterion) — via a conductor-owned append to `.epic-worklog.md` (append-only, so it can't clobber worker history) rather than a merge into `lead_guidance`.
- Replace the pointer rule: state what changed in at least one line instead of pointing at a file the worker already reads.
- Add: the `SendMessage`/`ListAgents` address is the session name, not the tmux window name assigned at spawn.
- Add: when not to nudge (a worker in `awaiting-results` with a live `check_cmd`), and `notify_when_idle` as the anti-polling alternative, with its three bounds (main-conversation only, this-machine only, one-shot).
- The three surviving rules (delivery timing, addressable-sessions-only, no permission laundering) are preserved byte-identical in `skills/bip-epic/SKILL.md` and in substance in `skills/bip-epic-poll/SKILL.md`.

Full rationale and the test-worker evidence (a `conductor_probe` key silently dropped across 10 status rewrites, a `SendMessage` failure against a tmux window name, etc.) are in #190.

Closes #190

## Test plan
- Verbatim-preservation check: the three kept bullets in `skills/bip-epic/SKILL.md` are byte-identical to `git show c19f9e2:skills/bip-epic/SKILL.md` (confirmed via exact substring match), and the old dual-write/pointer wording no longer appears in either skill file.
- Docs-only change; no build or test suite applies.

### DIFF
--- a/skills/bip-epic-poll/SKILL.md
+++ b/skills/bip-epic-poll/SKILL.md
@@ -124,17 +124,30 @@ the plan.
 **Correcting a live worker mid-flight**: `ListAgents`/`SendMessage`
 reach other local Claude tmux sessions and can push an immediate
 correction without waiting for the worker's next lead evaluation or
-killing its tmux window. `.epic-status.json`'s `lead_guidance` remains
-the durable, canonical instruction — any `SendMessage` correction MUST
-also be written to `lead_guidance` in the same step, so a compacted or
-restarted worker can reconstruct it from files alone. Delivery lands
-at the worker's next tool call, not instantly, and `SendMessage` only
+killing its tmux window. Send the correction directly — stating what
+changed in at least one line — rather than a bare pointer, which is a
+no-op for a worker that already reads the status file every step and
+strips the priority signal. The durable/transient test is the
+deliverable: any nudge that changes what the worker will produce
+(scope, target, artifact, gate criterion) MUST also get a durable
+record, written by the **conductor** appending a timestamped,
+attributed entry to `.epic-worklog.md` in the same step — not
+`.epic-status.json`, since that file is the worker's own
+continuously-rewritten object and a second writer there is
+unsynchronized and gives no tiebreak against `lead_guidance`. Facts
+that leave the deliverable unchanged (host load, a peer's timing, a
+dependency that just landed) may be message-only. Delivery lands at
+the worker's next tool call, not instantly, and `SendMessage` only
 reaches addressable Claude sessions — run `ListAgents` first to
-confirm, and fall back to the file-only correction (edit
-`lead_guidance`, wait for the worker's own loop) when it isn't
-addressable. Send a pointer ("updated .epic-status.json lead_guidance
-for iN — re-read it"), not the instruction payload, so nothing
-important lives only in the ephemeral channel.
+confirm (the address is the session name it reports, not the tmux
+window name assigned at spawn — sending to the window name can fail
+with `No agent named '<window>' is reachable.`), and fall back to the
+file-only correction (a `conductor_guidance` field or a `lead_notes`
+entry tagged `source: conductor` — never `lead_guidance`, wait for the
+worker's own loop) when it isn't addressable. Skip the nudge entirely
+for a worker in `awaiting-results` with a live `check_cmd`; use
+`notify_when_idle: true` instead of "tell me when this worker
+finishes" — main-conversation only, this-machine only, one-shot.
 
 ### Output structure
 
--- a/skills/bip-epic-poll/SKILL.md
+++ b/skills/bip-epic-poll/SKILL.md
@@ -124,30 +124,36 @@ the plan.
 **Correcting a live worker mid-flight**: `ListAgents`/`SendMessage`
 reach other local Claude tmux sessions and can push an immediate
 correction without waiting for the worker's next lead evaluation or
-killing its tmux window. Send the correction directly — stating what
-changed in at least one line — rather than a bare pointer, which is a
-no-op for a worker that already reads the status file every step and
-strips the priority signal. The durable/transient test is the
-deliverable: any nudge that changes what the worker will produce
-(scope, target, artifact, gate criterion) MUST also get a durable
-record, written by the **conductor** appending a timestamped,
-attributed entry to `.epic-worklog.md` in the same step — not
-`.epic-status.json`, since that file is the worker's own
+killing its tmux window.
+Send the correction directly — stating what changed in at least one
+line — rather than a bare pointer, which is a no-op for a worker that
+already reads the status file every step and strips the priority
+signal.
+If a correction is long enough that pointing at a file seems
+preferable, its home is the spawn prompt or the issue body, not a
+nudge.
+The durable/transient test is the deliverable: any nudge that changes
+what the worker will produce (scope, target, artifact, gate criterion)
+MUST also get a durable record, written by the **conductor** appending
+a timestamped, attributed entry to `.epic-worklog.md` in the same
+step — not `.epic-status.json`, since that file is the worker's own
 continuously-rewritten object and a second writer there is
-unsynchronized and gives no tiebreak against `lead_guidance`. Facts
-that leave the deliverable unchanged (host load, a peer's timing, a
-dependency that just landed) may be message-only. Delivery lands at
-the worker's next tool call, not instantly, and `SendMessage` only
-reaches addressable Claude sessions — run `ListAgents` first to
-confirm (the address is the session name it reports, not the tmux
-window name assigned at spawn — sending to the window name can fail
-with `No agent named '<window>' is reachable.`), and fall back to the
-file-only correction (a `conductor_guidance` field or a `lead_notes`
-entry tagged `source: conductor` — never `lead_guidance`, wait for the
-worker's own loop) when it isn't addressable. Skip the nudge entirely
-for a worker in `awaiting-results` with a live `check_cmd`; use
-`notify_when_idle: true` instead of "tell me when this worker
-finishes" — main-conversation only, this-machine only, one-shot.
+unsynchronized and gives no tiebreak against `lead_guidance`.
+Facts that leave the deliverable unchanged (host load, a peer's
+timing, a dependency that just landed) may be message-only.
+Delivery lands at the worker's next tool call, not instantly, and
+`SendMessage` only reaches addressable Claude sessions — run
+`ListAgents` first to confirm (the address is the session name it
+reports, not the tmux window name assigned at spawn — sending to the
+window name can fail with `No agent named '<window>' is reachable.`),
+and fall back to the file-only correction (a `conductor_guidance`
+field or a `lead_notes` entry tagged `source: conductor` — never
+`lead_guidance`, wait for the worker's own loop) when it isn't
+addressable.
+Skip the nudge entirely for a worker in `awaiting-results` with a live
+`check_cmd`; use `notify_when_idle: true` instead of "tell me when
+this worker finishes" — main-conversation only, this-machine only,
+one-shot.
 
 ### Output structure
 
--- a/skills/bip-epic/SKILL.md
+++ b/skills/bip-epic/SKILL.md
@@ -82,8 +82,10 @@ never replaces — the file-based correction path:
   windows on this machine, or connected cloud/Remote Control sessions)
   — never a plain shell, a remote SSH job, or a non-Claude compute node.
   Run `ListAgents` first to confirm the target session is actually
-  addressable; fall back to editing `.epic-status.json` and waiting for
-  the worker's own loop when it isn't. **The address is the session
+  addressable; fall back to the file-only correction (a
+  `conductor_guidance` field or a `lead_notes` entry tagged `source:
+  conductor` — never `lead_guidance`) and wait for the worker's own
+  loop when it isn't. **The address is the session
   name `ListAgents` reports, not the tmux window name the conductor
   assigned at spawn** — `SendMessage` to a window name can fail with
   `No agent named '<window>' is reachable.`
--- a/skills/bip-epic/SKILL.md
+++ b/skills/bip-epic/SKILL.md
@@ -50,11 +50,31 @@ let the conductor push an immediate correction to a live worker
 without killing and respawning its tmux window. This supplements —
 never replaces — the file-based correction path:
 
-- **`.epic-status.json`'s `lead_guidance` remains the durable, canonical
-  instruction.** Any correction sent via `SendMessage` MUST also be
-  written to `lead_guidance` in the same step, so a compacted or
-  restarted worker (or a fresh conductor cold-starting via `/bip-epic`)
-  can reconstruct the correction from files alone.
+- **Send the correction directly — that is the channel's purpose.**
+  No status-file write is a precondition for messaging. State the
+  change in at least one line; a bare pointer ("re-read
+  `lead_guidance`") is a no-op for a worker that already reads the
+  status file every step, and it strips the priority signal (drop
+  what you're doing vs. finish the current step). If a correction is
+  long enough that pointing at a file seems preferable, its home is
+  the spawn prompt or the issue body, not a nudge.
+- **The durable/transient test is the deliverable.** Any nudge that
+  changes what the worker will produce — scope, target, artifact, gate
+  criterion — is state that matters and MUST be recorded durably in
+  the same step as the message. Facts that leave the deliverable
+  unchanged (which host to use, a peer's timing, a dependency that
+  just landed) may be message-only.
+  - Record deliverable-changing corrections by having the **conductor**
+    append a timestamped, attributed entry to `.epic-worklog.md` in
+    the same step — not by writing `.epic-status.json`. Workers treat
+    that file as append-only and never edit previous entries, so a
+    conductor append cannot clobber worker history the way editing
+    `lead_guidance` can: it is a field every spawn prompt attributes to
+    the lead, so a second writer makes a conductor nudge
+    indistinguishable from a lead verdict, with no tiebreak when they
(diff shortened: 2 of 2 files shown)


matsen:
Checked this against #190's seven success criteria mechanically, including running the verbatim gate from the issue's own test plan. **Six pass cleanly, one is partial, plus one asymmetry and a convention suggestion.**

## Criteria results

| # | criterion | result |
|---|---|---|
| 1 | no status-write precondition for messaging | pass — and all three replaced wordings return zero hits |
| 2 | `.epic-worklog.md` named + non-lead-owned field | pass in both files |
| 3 | address is the `ListAgents` session name + failure text quoted | pass in both |
| 4 | when-not-to-nudge + `notify_when_idle` with its three bounds | pass in both |
| 5 | three kept rules survive | **pass** — all three bullets are still exact substrings of `c19f9e2` |
| 6 | durable/transient test stated in terms of the deliverable | pass in both |
| 7 | no bare pointers | **partial** — see below |

Criterion 1 and 5 verified with the issue's own commands:

```
$ for f in /tmp/k2.txt /tmp/k3.txt /tmp/k4.txt; do grep -qF -f "$f" skills/bip-epic/SKILL.md && echo "OK $f" || echo "DRIFT $f"; done
OK  /tmp/k2.txt
OK  /tmp/k3.txt
OK  /tmp/k4.txt

$ grep -nE "MUST also be written to .lead_guidance|not the instruction payload|rather than the instruction itself" skills/bip-epic/SKILL.md skills/bip-epic-poll/SKILL.md
(zero hits)
```

Criterion 5 passing is better than the issue deserved. #190's review flagged that criteria 3 and 5 would fight — 3 demands adding the address statement, 5 demands the kept bullets stay byte-identical — and proposed a licensed exception. This PR dissolves the conflict instead, by appending the address sentence *after* the bullet's final period so `grep -qF` on the original text still matches. No exception needed. Worth keeping that shape in mind for future edits to these bullets: append, don't interleave.

## 1. Criterion 7's third clause is missing from `bip-epic-poll` (blocking)

Criterion 7 requires both skills to say that a correction too long to send belongs in the spawn prompt or the issue body, not a nudge. `bip-epic/SKILL.md` has it; `bip-epic-poll/SKILL.md` does not:

```
$ for f in skills/bip-epic/SKILL.md skills/bip-epic-poll/SKILL.md; do
    printf "%-32s %s\n" "$f" "$(tr '\n' ' ' < $f | grep -coE 'spawn prompt or the issue body')"; done
skills/bip-epic/SKILL.md         1
skills/bip-epic-poll/SKILL.md    0
```

This clause is load-bearing, not decoration. Without it the poll paragraph says "send the correction directly, stating what changed in at least one line" with no upper bound — which invites exactly the failure that produced #190 in the first place: a conductor sending a live worker ~2,500 characters of standing protocol through a nudge channel, where it competes with the worker's own context and belongs in the spawn prompt instead.

Suggested insertion at `skills/bip-epic-poll/SKILL.md:130`, after "strips the priority signal.":

> If a correction is long enough that pointing at a file seems preferable, its home is the spawn prompt or the issue body, not a nudge.

## 2. The two files now disagree about the file-only fallback (non-blocking)

`bip-epic-poll:145-147` names the field explicitly — "a `conductor_guidance` field or a `lead_notes` entry tagged `source: conductor` — never `lead_guidance`". `bip-epic:85` still reads:

> addressable; fall back to editing `.epic-status.json` and waiting for the worker's own loop when it isn't.

Not wrong — the bullet above it forbids merging into `lead_guidance` — but it's the same instruction stated two different ways in two files, which is the drift shape this PR is cleaning up elsewhere. Since `bip-epic` is the fuller of the two, it should carry at least as much specificity as the summary. Suggested: mirror the poll file's parenthetical there.

## 3. Consider one sentence per line for these files

Not a criterion, and a broader question than this PR — but this PR is a good moment to raise it, because verifying it kept tripping over the current wrapping.

Both files are hard-wrapped at ~70 characters, so phrases break mid-sentence across lines. Three of my first grep checks returned false failures for this reason alone: `session name`, `main conversation`, and `durable/transient test is the deliverable` all *are* present but span a line break, so a single-line `grep` misses them. I had to re-run everything through `tr '\n' ' '` to get true answers — and one of those false negatives was a criterion I was about to report as unmet.

One sentence per line (semantic line breaks) fixes that and two other things:

- **Greps and checks work.** Each sentence is a complete line, so `grep` on a phrase behaves as expected — relevant here because #190's criteria are meant to be mechanically checkable, and right now checking them requires knowing to unwrap first.
- **Diffs become minimal and reviewable.** Editing one sentence touches one line instead of reflowing a whole paragraph. This PR's own diff is harder to read than it needs to be for exactly that reason: several hunks are reflow noise rather than content change.
- **Rendering is unchanged.** Markdown soft-wraps, and these files are consumed as prompt text where line breaks inside a paragraph don't matter either way.

It's the opposite of the convention `PROSE-DISCIPLINE.md` sets for GitHub issue bodies (one long line per paragraph, let GitHub wrap) — appropriately, since that rule exists so GitHub's renderer controls wrapping, whereas these files are read as source.

I'd suggest scoping it to the paragraphs this PR already rewrites — those lines are being touched anyway, so it's free — and filing a separate issue if it's wanted repo-wide across `skills/`. Converting all of `skills/` in this PR would bury the actual change in whitespace churn.

## Verdict

Ready to merge once the criterion-7 sentence is added. Items 2 and 3 are worth doing while you're in the file but neither blocks.

PR #199, opened 2026-08-28
Prune stale skill symlinks, make tuckin MEMORY.md optional, distinguish consumed spawn intent

### LINKED ISSUE #198: Prune stale skill symlinks on install, make the tuckin MEMORY.md step optional, and distinguish consumed spawn intent from pending
# Prune stale skill symlinks on install, make the tuckin MEMORY.md step optional, and distinguish consumed spawn intent from pending

🤖

Three independent fixes: `make install` removes symlinks whose source skill directory no longer exists; `/bip-conductor-tuckin` and `/bip-epic-tuckin` stop mandating a MEMORY.md write; and a consumed spawn-intent file stops looking identical to a queued one.

## Motivation

All three surfaced from one run — the first real use of `/bip-conductor-tuckin`, on the phyz fleet, immediately after #194's rename landed on a branch. None is hypothetical and each has a one-line-to-one-paragraph fix. Filing together because they were found together; they are independently actionable and can be split.

## 1. `make install` is additive-only, so renamed skills leave dangling symlinks

`symlink-skills` (`Makefile:20-27`) iterates `$(REPO_DIR)/skills/*/` and links each one. It never enumerates what is already in `~/.claude/skills/`, so a symlink whose source directory was renamed or deleted survives forever.

#194 renames five fleet skills from `bip-epic-*` to `bip-conductor-*`. Measured on this machine after a successful `make install` against that branch:

```
conductor skills linked:        7
stale bip-epic-* fleet links:   5   (bip-epic-spawn, -poll, -handoff, -recover, -prepare-reboot)
```

All five are dangling — their targets no longer exist. The new skills installed correctly; the old names were simply left behind.

This is worse than a cosmetic leftover. A dangling entry keeps a retired skill *name* present in the skills directory, so it can still be surfaced or attempted, and the failure mode when it is invoked is confusing rather than "no such skill". It also means every future rename silently accumulates another one.

Fix: after the link loop, remove entries in `~/.claude/skills/` that are symlinks and do not resolve. Prune only broken links — never anything that resolves — so hand-made or plugin-provided skills that do not come from this repo are untouched.

## 2. The tuckin skills mandate a MEMORY.md write, and its examples are the anti-pattern

`/bip-conductor-tuckin` Step 3 ("Update MEMORY.md (lightweight)") and the equivalent step in `/bip-epic-tuckin` instruct a write to the auto-memory directory's `MEMORY.md`. Three references in each skill, plus one in `/bip-epic` Step 1.

Two problems, and the second is the larger one.

**It is a MUST with no alternative.** Some setups deliberately do not use those files. The stated reason on this machine, from a global `CLAUDE.md`: the memory directory is keyed by working directory, so anything written there is invisible to other clones of the same repo and to the user. An agent following the step either violates that policy or silently skips a numbered step with no guidance on what to do instead. The directory is auto-created and usually exists, so the write succeeds and the conflict never surfaces.

**Its examples contradict #194's own rule.** Step 3 offers "host/cache quirks (which remote host had a warm cache for what, which host was mid-build)" as what to record. That is derived, transient state — exactly what `/bip-epic` Step 5 now forbids authoring: fleet state is derived, never written down, because it goes stale within a day with no mechanism to notice. The skill illustrates the practice its sibling skill rules out.

Measured against a real run: the first production use of `/bip-conductor-tuckin` produced five candidate fleet-level facts. Four were derived — which host held which build (stale within hours), local cache sizes (re-measurable with `du`), the open-issue count (`gh`), and which worker was blocked (visible in tmux). The fifth, that the nightly had gone green after sixteen red nights, was a genuine durable finding and was already recorded in its own issue. Nothing needed a new home.

Fix: replace the write instruction with a filter, then a route. The filter does the work:

1. **Is it derived?** Recomputable from `git`, `tmux`, `gh`, or the filesystem — record nothing, whatever the destination would have been. This is most of what a tuckin surfaces.
2. **Is it already recorded?** A finding that produced an issue, PR, test, or doc needs no second copy.
3. Only what survives both gets a destination: a durable repo-level fact belongs in a `CLAUDE.md`; a workflow rule belongs in a skill; a finding belongs in the test, doc, or issue it came from; and if none fit, it goes in the Step 4 report, which is where a user who has opted out of memory files will actually see it.

Replace the two host/cache examples, which fail gate 1.

## 3. A consumed spawn-intent file is indistinguishable from a queued one

`/bip-conductor-spawn` Step 5 says to delete the intent file after a successful launch. `/bip-conductor-tuckin` Step 2 says to list files "still waiting on execution" and report them so the next session knows what is queued. Nothing records that a file *was* consumed, so the two states are the same on disk.

Observed at tuckin: `.spawn-prompts/` held `spawn-2047.txt` and `spawn-2051.txt`, both already launched. Reported as pending, they invite a re-spawn of work that is already running or merged. Deleting them at tuckin instead is unattractive for two reasons: the directory is deliberately outside every clone's git, so deletion is unrecoverable; and one of the two belonged to an issue still open with a live session working from that brief.

So the conductor at tuckin has no correct move — report them and mislead, or delete them and destroy an authored artifact that may still be in use.

Fix, cheapest first: on successful launch, rather than deleting, record consumption in a way the next reader can see — a `consumed/` subdirectory, or a marker line appended by the conductor. Then tuckin's Step 2 can report queued and consumed separately, and the "delete after launch" instruction can be softened to "move aside", which is reversible. Whatever shape is chosen, `/bip-conductor-spawn` Step 5 and `/bip-conductor-tuckin` Step 2 must agree — they currently prescribe different treatments of the same file. Note `skills/lib/spawn-intent.sh` now owns intent-path resolution (#195/PR #196); the lifecycle marker should live alongside it rather than as a fourth inline convention.

## Files to modify

- `Makefile` — the `symlink-skills` target, adding a prune of broken links
- `skills/bip-conductor-tuckin/SKILL.md` — Step 2 (consumed vs queued), Step 3 (filter-then-route in place of the MEMORY.md write)
- `skills/bip-epic-tuckin/SKILL.md` — the equivalent MEMORY.md step
- `skills/bip-conductor-spawn/SKILL.md` — Step 5's post-launch treatment of the intent file
- `skills/lib/spawn-intent.sh` — the consumed/queued marker, if it lands there
- `skills/bip-epic/SKILL.md` — Step 1's MEMORY.md read, if it should be conditional too

## Test plan

- **Prune**: create a symlink in `~/.claude/skills/` pointing at a nonexistent directory plus one pointing at a real non-repo directory, run `make install`, and assert the broken one is gone and the resolving one survives. The second half is what stops an over-eager prune from deleting a user's own skills.
- **Rename round-trip**: rename a skill directory in the repo, run `make install`, and assert no dangling entry remains under the old name.
- **MEMORY.md**: run the tuckin step with the memory directory absent, and with it present but opted out, and confirm the same surviving content reaches the Step 4 report in both cases — and that derived facts reach neither.
- **Intent lifecycle**: launch from an intent file, then run tuckin, and assert the file is reported as consumed rather than queued, and that nothing was irrecoverably deleted.

## Success criteria

1. After `make install` against a branch that renames a skill, `~/.claude/skills/` contains no broken symlinks, and every symlink that resolved before still resolves.
2. Neither tuckin skill instructs an unconditional MEMORY.md write. Both apply the derived/already-recorded filter before routing anything, state where surviving content goes when the project does not use memory files, and no longer offer derived host/cache state as an example of what to record.
3. `/bip-conductor-spawn` and `/bip-conductor-tuckin` prescribe the same treatment of an intent file after launch, and a consumed file is distinguishable from a queued one without reading its contents.

## Scope boundaries

**In scope**: the three fixes above.

**Out of scope**:

- Extracting `clone_root`/intent resolution into a tested sourceable script — that landed as #195 / PR #196 (`skills/lib/spawn-intent.sh`). This issue is about the intent file's *lifecycle*, not how its path is resolved.
- One-sentence-per-line reflow of the touched skills — that is #192, which covers all 41 files in one pass.
- Whether the auto-memory mechanism is a good idea in general. This only asks that skills not mandate it.

## References

- #194 — the rename that exposed finding 1, and the skills findings 2 and 3 live in.
- #195 / PR #196 — `skills/lib/spawn-intent.sh`, which finding 3 should build on rather than duplicate.
- #192 — the reflow pass, explicitly out of scope here.

matsen:
## Summary

Three independent, small fixes surfaced from the first real use of `/bip-conductor-tuckin` (phyz fleet, right after #194's skill rename):

- `make install`'s `symlink-skills` and `symlink-agents` targets now prune broken symlinks under `~/.claude/skills/` and `~/.claude/agents/` after linking, so a renamed or deleted skill/agent doesn't leave a dangling entry under its old name. Only symlinks that fail to resolve are removed — real directories and symlinks that still resolve (e.g. a hand-made or plugin skill) are untouched.
- `/bip-conductor-tuckin` and `/bip-epic-tuckin` no longer mandate a MEMORY.md write. Both now apply a two-gate filter (is it derived from `git`/`tmux`/`gh`/filesystem? does it already have a home — an issue, PR, EPIC body, test, or doc?) before routing whatever survives to a `CLAUDE.md`, a skill, or the tuckin report itself — the destination a project that doesn't use auto-memory files will actually see. `/bip-epic`'s MEMORY.md read is now conditional on the project using auto-memory at all.
- `skills/lib/spawn-intent.sh` gains `mark_spawn_intent_consumed()`, which moves a launched intent file into `.spawn-prompts/consumed/` instead of deleting it (the directory lives outside every clone's git, so deletion there is unrecoverable). `/bip-conductor-spawn` Step 6 calls it after a successful launch; `/bip-conductor-tuckin` Step 2 now reports queued and consumed spawn intent as separate groups, using `find -type f` so the `consumed/` directory itself is never misread as a queued entry. `/bip-conductor` and `/bip-conductor-poll`, the other two readers of `.spawn-prompts/`, are updated to ignore `consumed/` too.

Closes #198

**Migration note**: this convention is forward-looking. Any `.spawn-prompts/` directory that already holds top-level files for intents launched *before* this PR merges (e.g. the phyz fleet's `spawn-2047.txt`, `spawn-2051.txt`) predates `mark_spawn_intent_consumed()` and needs a one-time manual move (`mkdir -p consumed && mv <file> consumed/`) on that machine — this PR does not migrate existing fleets.

## Test plan

- `tests/integration/spawn_intent_test.go`: `TestMarkSpawnIntentConsumed` (moves the file, leaves nothing at the original path, confirms `find_spawn_intent` no longer reports it as queued) and `TestMarkSpawnIntentConsumedMissingSourceFailsFast` (a missing source errors rather than silently no-op-ing, so a failed launch can't look consumed). Existing `spawn-intent.sh` tests still pass.
- Manually verified the `make install` prune for both `symlink-skills` and `symlink-agents`: created a broken symlink, a resolving symlink (standing in for a hand-made skill/agent), and a plain directory under fake `$HOME/.claude/{skills,agents}`, ran the targets, confirmed only the broken symlink was removed.
- `go build`, `go vet`, `gofmt -l .`, and `go test ./...` all clean.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,12 @@ symlink-skills:
 		fi; \
 	done
 	@rm -rf ~/.claude/skills/lib && ln -s $(REPO_DIR)/skills/lib ~/.claude/skills/lib
+	@for l in ~/.claude/skills/*; do \
+		if [ -L "$$l" ] && [ ! -e "$$l" ]; then \
+			echo "Pruning stale symlink: $$l"; \
+			rm -f "$$l"; \
+		fi; \
+	done
 	@echo "Symlinked skills to ~/.claude/skills/"
 
 symlink-statusline:
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,12 @@ symlink-agents:
 	@for f in $(REPO_DIR)/agents/*.md; do \
 		ln -sf "$$f" ~/.claude/agents/$$(basename "$$f"); \
 	done
+	@for l in ~/.claude/agents/*; do \
+		if [ -L "$$l" ] && [ ! -e "$$l" ]; then \
+			echo "Pruning stale symlink: $$l"; \
+			rm -f "$$l"; \
+		fi; \
+	done
 	@echo "Symlinked agents to ~/.claude/agents/"
 
 symlink-skills:
--- a/skills/bip-conductor-poll/SKILL.md
+++ b/skills/bip-conductor-poll/SKILL.md
@@ -89,11 +89,13 @@ pattern in `SUBAGENT-SCAN.md` (bipartite repo root). Brief:
 >   issue-lead comments, slot phase changes
 > - `active_items`: per active slot — clone, issue, phase,
 >   stop_reason, lead assessment (one line each)
-> - `action_candidates`: pending spawn intent in `$CLONE_ROOT/.spawn-prompts/`
->   (either `<N>.md` or `spawn-<N>.txt` — check both) waiting to be
->   executed; merged PRs that should trigger slot cleanup. (Deciding *which* open
->   issues are ready to spawn is `/bip-epic`'s call, not this poll's —
->   report raw signal, not a readiness verdict.)
+> - `action_candidates`: pending spawn intent directly in
+>   `$CLONE_ROOT/.spawn-prompts/` (either `<N>.md` or `spawn-<N>.txt` —
+>   check both) waiting to be executed — ignore anything under
+>   `.spawn-prompts/consumed/`, that's already-launched intent, not
+>   pending; merged PRs that should trigger slot cleanup. (Deciding
+>   *which* open issues are ready to spawn is `/bip-epic`'s call, not
+>   this poll's — report raw signal, not a readiness verdict.)
 > - `surprises`: `needs-human`/`completed` slots, stale status
 >   files, contradictions, `RECOMMEND DEEPER LOOK` flags
 
--- a/skills/bip-conductor-spawn/SKILL.md
+++ b/skills/bip-conductor-spawn/SKILL.md
@@ -58,8 +58,10 @@ narrow this check to one pattern.
   says. Your job is to check it against live fleet state (Step 2b) and
   append fleet facts it structurally couldn't know: which host/clone is
   actually free, a concurrent worker editing an overlapping file, a
-  build running on a target remote host. Delete the intent file after a
-  successful launch (Step 5) — it has done its job.
+  build running on a target remote host. Mark the intent file consumed
+  after a successful launch (Step 6) — it has done its job, but the
+  directory lives outside git, so this is a move to `consumed/`, not a
+  delete (see Step 6).
 - **No intent file** (conductor-initiated respawn, routine maintenance,
   quick fix with no upstream epic session): compose the prompt from the
   issue directly, same as before. This is the escape hatch, not the
@@ -552,9 +554,17 @@ Always go through `bip spawn` which handles the full lifecycle correctly.
 ### Step 6: Confirm
 
 If launch succeeded and a `.spawn-prompts/` intent file (`$INTENT` from
-Step 3) was consumed, delete it now — its job is done, and leaving it
-behind is exactly the kind of finished-but-undeleted artifact that
-accretes across a fleet.
+Step 3) was used, mark it consumed now — don't delete it. The
+directory lives outside every clone's git, so deletion there is
+unrecoverable, and a still-open issue may have another live session
+referencing the same brief. Moving it aside is reversible and lets
+`/bip-conductor-tuckin` Step 2 report it as consumed rather than
+queued:
+
+```bash
+source "$(dirname "<this-skill's-base-directory>")/lib/spawn-intent.sh"
+mark_spawn_intent_consumed "$INTENT"
+```
 
 Report to the user:
 - Which clone was spawned
--- a/skills/bip-conductor-tuckin/SKILL.md
+++ b/skills/bip-conductor-tuckin/SKILL.md
@@ -62,9 +62,12 @@ List `$CLONE_ROOT/.spawn-prompts/` and report two groups separately —
 these are the epic's authored intent, not conductor state, so don't
 touch their contents either way:
 
-- **Queued** — files directly in `.spawn-prompts/` (either `<N>.md` or
-  `spawn-<N>.txt` — check both, the directory holds both live
-  conventions). Still waiting on execution.
+- **Queued** — regular files directly in `.spawn-prompts/`:
+  `find "$CLONE_ROOT/.spawn-prompts" -maxdepth 1 -type f \( -name
+  '*.md' -o -name 'spawn-*.txt' \)` (both naming conventions live
+  there). `-type f` matters: a plain `ls` also lists the `consumed/`
+  directory itself as an entry, which is not a queued intent file.
+  Still waiting on execution.
 - **Consumed** — files under `.spawn-prompts/consumed/`. Already
   launched by `/bip-conductor-spawn` Step 6.
 
--- a/skills/bip-conductor-tuckin/SKILL.md
+++ b/skills/bip-conductor-tuckin/SKILL.md
@@ -56,26 +56,47 @@ Update slots the conductor has direct knowledge about. This includes:
 Do not guess status for slots with active sessions that may have
 progressed beyond what the conductor last observed.
 
-### Step 2: Note pending spawn intent
+### Step 2: Note queued and consumed spawn intent
 
-List any `$CLONE_ROOT/.spawn-prompts/` files still waiting on
-execution (either `<N>.md` or `spawn-<N>.txt` — check both, the
-directory holds both live conventions) — these are the epic's authored
-intent, not conductor state, so don't touch their contents, just
-confirm they're still there and
-report them so the next conductor session (or the user) knows what's
-queued.
+List `$CLONE_ROOT/.spawn-prompts/` and report two groups separately —
+these are the epic's authored intent, not conductor state, so don't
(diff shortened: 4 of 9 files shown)


matsen:
Verified all three by execution rather than diff-reading.

**Prune** — sandboxed a fake `~/.claude/skills` with three entries and ran the exact loop: a broken symlink, a symlink resolving to a real directory (standing in for a hand-made skill), and a plain directory. Only the broken one was removed. The plain-directory case wasn't in the issue's test plan and it's the one an over-eager `find -type l -delete` would have caught wrongly, so good that it survives.

**Intent lifecycle** — `mark_spawn_intent_consumed()` moves rather than deletes, `consumed/` is a sibling so `find_spawn_intent`'s explicit-name glob can't see it, and `go test ./tests/integration -run SpawnIntent` passes 7/7 including `TestMarkSpawnIntentConsumedMissingSourceFailsFast`, which I hadn't asked for and which is the right instinct — a silent no-op on a missing source would make a failed launch look consumed.

**Filter** — Step 3 now names the anti-pattern examples explicitly (warm cache, mid-build, cache sizes, open-issue counts, which worker was blocked) as things to record *nowhere*. Naming them beats stating the principle abstractly, since those were the skill's own former examples.

## Two gaps

**1. The two live consumed files aren't migrated.** `~/re/pz/.spawn-prompts/` still holds `spawn-2047.txt` and `spawn-2051.txt` at top level. Both were launched — they're the files that motivated finding 3. After this merges, tuckin's Step 2 will report both as **queued**, which is exactly the failure the PR fixes, for the two files that prompted it. #2047 is closed and merged; #2051 is open with a live session still working from that brief.

Not a code change — either a one-time `mkdir -p consumed && mv` on this machine, or a line in the PR/issue saying existing top-level files predate the convention and need moving once. I'd rather it be stated than discovered by the next conductor.

**2. Step 2's listing should say files, not entries.** "Queued — files directly in `.spawn-prompts/`" is prose, and `consumed/` is now itself an entry in that directory. A naive `ls .spawn-prompts/` lists it alongside the queued files and invites reporting a directory as a queued intent. `find_spawn_intent` is unaffected (explicit-name glob), so this is only the human-readable listing. Suggest `find "$CLONE_ROOT/.spawn-prompts" -maxdepth 1 -type f` or an explicit "directories are not intent".

## One structural note, bigger than this PR

Worth recording here but not fixing here. This PR makes Step 3 correctly filter out nearly everything — which surfaces that tuckin's remaining real work is **reconciliation, not persistence**. In the run that produced #198, the only durable output was repairing one slot's `.epic-status.json`, which had read `phase: coding` since 2026-08-25 while the issue closed and the clone returned to `main`. Two days of drift, caught only because a context reset happened to occur.

`/bip-conductor`'s own scan already *detects* that (it classifies slots `stale` and surfaces `action_candidates`), but the *repair* exists only in tuckin. So drift is re-noticed every cold start and fixed only at reset. That's a placement question about the reconcile duty, not a defect in this PR, and it wants its own issue.

Approving from my side once gap 1 is stated somewhere durable.

matsen:
Thanks for the execution-verified review. Both gaps addressed:

1. **Migration**: added a note to the PR body — this convention is forward-looking, existing top-level files (including phyz's `spawn-2047.txt`/`spawn-2051.txt`) predate it and need a one-time manual `mkdir -p consumed && mv` on that machine. Not a code fix, so didn't invent one.
2. **Step 2 listing**: fixed — now `find "$CLONE_ROOT/.spawn-prompts" -maxdepth 1 -type f ...` instead of unqualified prose, so `consumed/` itself can't be misread as a queued entry.

On the structural note (reconcile vs. persist): agree it doesn't belong in this PR, and I don't want to resolve it unilaterally either — it changes where a repair duty lives across three skills you didn't just review, that's a call for the person who actually runs this fleet day to day. Passing your three questions to the user rather than answering for them; will follow up with a single issue once that's settled, per your preference for one over two.

PR #206, opened 2026-08-28
Push completion notifications from worker to conductor and conductor to epic

### LINKED ISSUE #205: Push completion notifications from worker to conductor and from conductor to epic
# Push completion notifications from worker to conductor and from conductor to epic

🤖 Replace the unused phone/bell notification with a `SendMessage` push in both completion directions — worker→conductor and conductor→epic — reusing the `ListAgents`/`SendMessage` mechanics already built for corrections and spawn intent, with the existing file/log/poll path kept as the fallback.

## Problem

The EPIC fleet already has a push channel for *corrections* — `/bip-epic` drafts a correction, the conductor `SendMessage`s it to the worker directly (skills/bip-conductor/SKILL.md "Correcting a live worker") — and for *spawn intent* — `/bip-epic` writes `.spawn-prompts/`, the conductor picks it up. Completion signals run the opposite direction and have no equivalent push:

1. **Worker → conductor**: a worker reaching `needs-human`/`completed` only rings the terminal bell and sends an `ntfy` push (skills/bip-conductor-spawn/SKILL.md, PHONE NOTIFICATION step). That push targets the human directly, and it's dead weight in practice — `ntfy_topic` was never configured, so the `curl` is a silent no-op, and the bell alone is just noise. The conductor learns of the completion only by polling `.epic-status.json`/`.epic-notifications.log` (`/bip-conductor-poll`, `bip epic watch` in cmd/bip/epic_watch.go) — a conductor mid-conversation with the user has no reason to check either until its next natural break.
2. **Conductor → epic**: this hop doesn't exist at all. `/bip-epic` discovers merged PRs and closed issues only through its own independent `gh` polling (skills/bip-epic/SKILL.md Step 2), never through anything the conductor already observed.

The same dead bell/`ntfy` pattern also appears in `skills/bip-ms-poll/SKILL.md` (manuscript polling, unrelated to the EPIC fleet) — same unconfigured `ntfy_topic`, same no-op.

## Proposed implementation

1. **Worker → conductor** (skills/bip-conductor-spawn/SKILL.md, PHONE NOTIFICATION step): replace the bell/`ntfy` block with: run `ListAgents` for a conductor session; if one is addressable, `SendMessage` it a one-line notification (issue, phase, one-line summary). `.epic-status.json` is written regardless, so `/bip-conductor-poll` and `bip epic watch` remain the fallback when no conductor is addressable.
2. **Conductor → epic**: two sites observe a `needs-human`/`completed` transition and both need the same push added — `skills/bip-conductor-poll/SKILL.md`'s "Flag needs-human and completed" section (drop the bell/`ntfy` block there too) and `skills/bip-conductor/SKILL.md` Step 7's "When a transition arrives showing `needs-human` or `completed`, the conductor should react immediately" line (add the push as part of that reaction, alongside reading the slot's status and checking lead guidance). In both places: run `ListAgents` for an epic session; if one is addressable, `SendMessage` it the issue number and phase so `/bip-epic` can re-poll and update the EPIC body without waiting for its own cadence. If none is addressable, no change — `/bip-epic`'s independent `gh` polling remains the fallback path.
3. **skills/bip-ms-poll/SKILL.md**: remove the same bell/`ntfy` block (the "Ring the terminal bell and send a phone notification if a major new result arrives" step). No replacement needed — this skill has no conductor/epic split to push into; the existing report-to-the-user-in-conversation behavior already covers it.

Neither fleet hop is required for correctness — both are fallback-safe, since the file-based state (`.epic-status.json`, `.epic-notifications.log`, GitHub itself) remains the durable source of truth. This is strictly a latency improvement for the case where the sessions are already live, replacing a notification path that was never actually reaching anyone.

## Files to modify

- `skills/bip-conductor-spawn/SKILL.md` — PHONE NOTIFICATION step: replace bell/`ntfy` with the `ListAgents`/`SendMessage` push to the conductor.
- `skills/bip-conductor-poll/SKILL.md` — "Flag needs-human and completed" section: replace bell/`ntfy` with the `ListAgents`/`SendMessage` push to the epic session.
- `skills/bip-conductor/SKILL.md` — Step 7 ("When a transition arrives showing `needs-human` or `completed`, the conductor should react immediately"): add the same `ListAgents`/`SendMessage` push to the epic session as part of that reaction. Also update the Conventions section (the "SendMessage as a nudge channel" home) to note the completion-push direction alongside the existing correction-push documentation, since both now share the same addressability caveats.
- `skills/bip-ms-poll/SKILL.md` — remove the bell/`ntfy` block; no replacement.

## Test plan

No automated test — this is a skill/prompt change, not code. Verify by hand in a live fleet session:
1. Start a conductor and an epic session (`ListAgents` shows both addressable) plus a worker slot.
2. Drive the worker to `needs-human` or `completed`. Confirm the conductor session receives a `SendMessage` without needing to run `/bip-conductor-poll`, and confirm no bell/`curl` fires.
3. Confirm the epic session receives a follow-up `SendMessage` from the conductor and, on its next turn, re-polls and updates the relevant EPIC body.
4. Kill the epic session (make it non-addressable) and repeat step 2 — confirm the conductor falls back cleanly (no error, no retry loop) and the completion is still visible via `/bip-conductor-poll`.

## Success criteria

1. A worker reaching `needs-human`/`completed` with an addressable conductor session delivers a `SendMessage` to it and no longer rings the bell or calls `ntfy`.
2. A conductor observing a slot reach `needs-human`/`completed` with an addressable epic session delivers a `SendMessage` to it.
3. Neither push is a hard dependency: with no addressable target, existing poll/log/`gh`-based discovery is unchanged.
4. `skills/bip-ms-poll/SKILL.md` no longer contains a bell/`ntfy` block.

## Scope boundaries

**In scope**: the two notification pushes described above, removing the bell/`ntfy` blocks from all three skills that carry it, plus the SKILL.md documentation update noting the completion pushes share the corrections channel's addressability caveats.

**Out of scope**:
- The correction-delivery mechanics themselves (already built, in skills/bip-conductor/SKILL.md "Correcting a live worker" and skills/bip-epic/SKILL.md Step 7) — this issue only adds a second use of the same channel, in the opposite direction.
- Any change to `bip epic watch` or `.epic-notifications.log` — they remain the fallback path.
- `ntfy_topic` itself in `~/.config/bip/config.yml` — it's a personal config value, not something this repo manages.

## References

- skills/bip-conductor-spawn/SKILL.md (PHONE NOTIFICATION)
- skills/bip-conductor-poll/SKILL.md ("Flag needs-human and completed", "Correcting a live worker mid-flight")
- skills/bip-conductor/SKILL.md ("Correcting a live worker: SendMessage as a nudge channel — execution half")
- skills/bip-epic/SKILL.md (Step 2, Step 7)
- skills/bip-ms-poll/SKILL.md
- cmd/bip/epic_watch.go

matsen:
## Summary
Workers and the conductor push completion signals (`needs-human`/`completed`) directly to an addressable peer session instead of relying solely on file-based polling — replacing a dead bell/`ntfy` block that never reached anyone (`ntfy_topic` was never configured in `~/.config/bip/config.yml`).

- **Worker → conductor**: a worker reaching `needs-human`/`completed` reads `$CLONE_ROOT/.conductor-session` and `SendMessage`s that exact address (`skills/bip-conductor-spawn/SKILL.md`).
- **Conductor → epic**: the conductor does the same into `$CLONE_ROOT/.epic-session` on observing the same transition (`skills/bip-conductor-poll/SKILL.md`, `skills/bip-conductor/SKILL.md` Step 7).
- Each side **self-registers its own address** rather than the receiver guessing it. Session names derive from working directory, so sessions sharing a clone root share a name prefix, and a display name can drift mid-session — "run `ListAgents`, pick the addressable row" only works when a human is watching the pick, which is how the existing correction channel gets away with that wording. A completion push fires unsupervised, so guessing risks silently messaging the wrong live session. The conductor writes `.conductor-session` at cold start and every poll cycle; `/bip-epic` writes `.epic-session` the same way. This is documented in `skills/bip-conductor/SKILL.md`'s Conventions section ("Completion pushes").
- Neither push is a hard dependency: a missing registration file or a failed send (stale address) falls back silently to the existing `.epic-status.json` / `.epic-notifications.log` / `gh`-polling path.
- Removed the same dead bell/`ntfy` block from `skills/bip-ms-poll/SKILL.md` (manuscript polling) — no replacement, since that skill has no conductor/epic split to push into.

Closes #205

## Test plan
No automated test — this is a skill/prompt change. Verify by hand in a live fleet session:
- Confirm `$CLONE_ROOT/.conductor-session` and `.epic-session` are written at cold start and refreshed on each poll cycle.
- Drive a worker to `needs-human`/`completed` with an addressable conductor session; confirm the conductor receives the push and no bell/`curl` fires.
- Point a registration file at a session that has since exited (or delete it) and repeat; confirm the sender falls back silently with no error and no retry loop.
- With multiple sessions sharing a clone root (a name-collision setup), confirm the push still lands on the correct session, since the address comes from the registration file rather than from scanning `ListAgents`.

### DIFF
--- a/skills/bip-conductor-poll/SKILL.md
+++ b/skills/bip-conductor-poll/SKILL.md
@@ -85,12 +85,8 @@ This tells the conductor what the workers are doing without having to read full 
 
 **Flag needs-human and completed** — if any clone has `phase: "needs-human"` (or legacy `blocked`) or `phase: "completed"`, highlight it prominently.
 These require conductor attention.
-**Ring the terminal bell and send a phone notification** so the user notices even if away:
-```bash
-printf '\a'
-NTFY_TOPIC=$(grep ntfy_topic ~/.config/bip/config.yml | awk '{print $2}')
-[ -n "$NTFY_TOPIC" ] && curl -s -H "Title: bip epic" -d "<clone> <phase> (<issue>)" "ntfy.sh/$NTFY_TOPIC" > /dev/null
-```
+Run `ListAgents` for an addressable epic session and, if one exists, `SendMessage` it the issue number and phase so `/bip-epic` can re-poll and update the EPIC body without waiting for its own cadence.
+If no epic session is addressable, this is a no-op — `/bip-epic`'s independent `gh` polling remains the fallback path.
 
 **Only report active clones** — clones with a tmux window that are actually doing something.
 Don't list completed or idle clones; that's noise.
--- a/skills/bip-conductor-poll/SKILL.md
+++ b/skills/bip-conductor-poll/SKILL.md
@@ -23,7 +23,11 @@ For periodic auto-polling: `/loop 10m /bip-conductor-poll`
 
 ## What to check
 
-Two things stay in the primary because they're cheap and structured:
+Three things stay in the primary because they're cheap and structured:
+
+**Refresh the completion-push address** — resolve `CLONE_ROOT` and rewrite `$CLONE_ROOT/.conductor-session` with this session's current `ListAgents` name.
+This is the file workers read to push a `needs-human`/`completed` notification without guessing among `ListAgents` rows — see `/bip-conductor`'s Conventions section ("Completion pushes") for why.
+Cheap (one `ListAgents` self-lookup, one file write) and bounds how stale the address can get between conductor cold starts.
 
 **Notifications log tail** — if `bip epic watch` is running, it appends one JSONL line per phase transition to `.epic-notifications.log` in the conductor cwd.
 The state file `/tmp/.epic-poll-last-read` records the previous poll time as Unix seconds; pass the elapsed window to `--since`:
@@ -85,8 +89,8 @@ This tells the conductor what the workers are doing without having to read full 
 
 **Flag needs-human and completed** — if any clone has `phase: "needs-human"` (or legacy `blocked`) or `phase: "completed"`, highlight it prominently.
 These require conductor attention.
-Run `ListAgents` for an addressable epic session and, if one exists, `SendMessage` it the issue number and phase so `/bip-epic` can re-poll and update the EPIC body without waiting for its own cadence.
-If no epic session is addressable, this is a no-op — `/bip-epic`'s independent `gh` polling remains the fallback path.
+Read `$CLONE_ROOT/.epic-session` and, if present, `SendMessage` that address the issue number and phase so `/bip-epic` can re-poll and update the EPIC body without waiting for its own cadence.
+If the file is absent or the send fails, this is a no-op — `/bip-epic`'s independent `gh` polling remains the fallback path.
 
 **Only report active clones** — clones with a tmux window that are actually doing something.
 Don't list completed or idle clones; that's noise.
--- a/skills/bip-conductor-spawn/SKILL.md
+++ b/skills/bip-conductor-spawn/SKILL.md
@@ -250,13 +250,12 @@ If you launch a long-running experiment:
 4. After 3 consecutive check failures, set stop_reason to
    mechanical-blocker and invoke the lead
 
-PHONE NOTIFICATION — When you set phase to needs-human or completed,
-ring the terminal bell and send a push notification so the user notices:
-```bash
-printf '\a'
-NTFY_TOPIC=$(grep ntfy_topic ~/.config/bip/config.yml | awk '{print $2}')
-[ -n "$NTFY_TOPIC" ] && curl -s -H "Title: bip epic" -d "#N <phase>: <one-line summary>" "ntfy.sh/$NTFY_TOPIC" > /dev/null
-```
+PUSH NOTIFICATION — When you set phase to needs-human or completed, run
+`ListAgents` for an addressable conductor session and, if one exists,
+`SendMessage` it a one-line notification (issue number, phase, one-line
+summary) so the conductor doesn't have to wait for its next poll cycle.
+`.epic-status.json` is written regardless, so `/bip-conductor-poll` and
+`bip epic watch` remain the fallback when no conductor is addressable.
 Do this EVERY time you write needs-human or completed to .epic-status.json.
 
 STOPPING POINTS — When you reach a natural stopping point:
--- a/skills/bip-conductor-spawn/SKILL.md
+++ b/skills/bip-conductor-spawn/SKILL.md
@@ -250,13 +250,20 @@ If you launch a long-running experiment:
 4. After 3 consecutive check failures, set stop_reason to
    mechanical-blocker and invoke the lead
 
-PUSH NOTIFICATION — When you set phase to needs-human or completed, run
-`ListAgents` for an addressable conductor session and, if one exists,
-`SendMessage` it a one-line notification (issue number, phase, one-line
-summary) so the conductor doesn't have to wait for its next poll cycle.
-`.epic-status.json` is written regardless, so `/bip-conductor-poll` and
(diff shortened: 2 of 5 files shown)


matsen:
Exercised `21db6aa` in a live fleet — four active workers, six colliding `phyz-*` session names. Both tests below were actually executed, and test 2 independently by both of us with byte-identical output.

## The design is right

Each role writes only **its own** name, taken from `ListAgents`' "This session is ..." row; the reader takes an exact address and never disambiguates; a missing file or failed send falls through silently with no `ListAgents` substitute scan. That last clause is what structurally eliminates the single-wrong-candidate case — a worker in a clone finding exactly one same-prefix session that happens to be the wrong role. Nobody scans, so there is nothing to get wrong.

## Test 1 — pathological name round-trip: PASS

I performed the conductor's Step 1 self-registration for real. `/home/matsen/re/pz/.conductor-session` received my actual current name, which is not an identifier:

```
we've moved from a /bip-epic focus to a recent-issue focus. let's look at the recent issues, with th (Branch)
```

110 bytes of prose, containing an apostrophe, forward slashes, and parentheses. That is genuinely what `ListAgents` reports for this session — the display name derived from the user's first message and has mutated since (see below).

@bipartite-bf read the file and sent to that string verbatim. Delivered, no stripping required. Two things that confirms:

- `(Branch)` is **part of the address**, not a display suffix to trim. Worth knowing, because it looks trimmable.
- The documented path is Read-tool → `SendMessage`, never `cat` + shell interpolation, so the quoting hazard that name would otherwise pose does not apply. That is a property of the instructions as written, and it would break the moment someone "helpfully" rewrites the read side as a bash one-liner. Worth a comment at the read site saying so.

## Test 2 — stale address: PASS, clean and distinguishable

Name drift is demonstrated, not hypothetical: earlier today this same session (ref `578c94`) was `phyz-1a`, and that name is now dead. Sending to it returns:

```json
{"success":false,"message":"No agent named 'phyz-1a' is reachable. Did you mean: phyz-ca, phyz-4c, phyz-6f?\nUse ListAgents to see everyone you can message."}
```

We ran this independently and got identical output. So a drifted address fails **explicitly** — `success:false`, plus a "not reachable" message — rather than silently succeeding into nothing or resolving to a live session by prefix. A caller can branch on the tool's own `success` field without parsing text, which makes the accepted-residual-risk paragraph honest: the failure mode is detectable.

## One new hazard the test surfaced

That error carries a fuzzy-match suggestion list: **"Did you mean: phyz-ca, phyz-4c, phyz-6f?"** Those are live sessions, none of them the conductor. That prompt is written for a human. An agent reading a failed `SendMessage` inside an unsupervised push loop might reasonably try a suggestion — which reintroduces exactly the mis-resolution this design eliminates, at the one moment the guard has already fired.

Suggest an explicit line in the Conventions "Completion pushes" section: on a failed send, do not act on the tool's suggestion list; any failure means not-addressable, fall back, do not retry against a substitute. The existing text says "never retry-loop or fall back to scanning `ListAgents` for a substitute", which covers scanning but not the suggestion handed to you unasked.

## Verdict

Approve with that one line added. Everything the earlier rounds flagged is closed, and closed in the right place — the fix moved from "guess better" to "don't guess", which is the correct shape.

Non-blocking observation for whoever owns session naming rather than this PR: a display name that is 110 bytes of the user's prose and mutates mid-session is a rough edge, and this design now has to tolerate it. It does tolerate it, which is the point — but the tolerance is load-bearing, not incidental.

PR #208, opened 2026-08-28
Harden the epic/conductor seam: six interface rules, durable decisions file

### LINKED ISSUE #207: Harden the epic/conductor seam: six interface rules, with decisions recorded to a durable fleet file
# Harden the epic/conductor seam: six interface rules, with decisions recorded to a durable fleet file

🤖

`bip-epic` and `bip-conductor` gain six interface rules:

- A conductor→epic decision-relay channel, with every relay marked `PROVISIONAL` or `FINAL` (`bip-conductor/SKILL.md`, new subsection after "Completion pushes").
- An EPIC body may record a user decision only against a `FINAL` shape (`bip-epic/SKILL.md:140`).
- Step 4 splits into 4a dependency-direction and 4b file-overlap, with separate completion criteria (`bip-epic/SKILL.md:119`).
- Group A scanners dispatch after a conductor handshake, scoped to the gap (`bip-epic/SKILL.md:57`).
- Worker findings forward verbatim and attributed, with any conductor reading marked as the conductor's own (`bip-conductor/SKILL.md`, new subsection).
- Both the `FINAL` relays and the negative list — decisions already taken *against* an action, with reasons — append to a durable fleet-level file, `.epic-decisions.md` in the conductor cwd.

## Motivation

`ba06e87` divided `bip-epic` into a topic-scoped epic and a fleet-scoped conductor. One two-session run of that split (EPIC `matsengrp/phyz`#369, 9 slots) produced three seam errors — none inside either role's own domain:

- **Collision missed.** The epic cleared `phyz`#2087 and #2088 as dependency-independent and wrote "No collisions" into a spawn brief; both edit `experiments/2026-08-27-2051-heavy-v2-stated-estimator/scripts/rescore_cell.py`.
- **Finding over-read.** The conductor forwarded a worker's matrix-provenance finding as undercutting `phyz`#2098's matched-model premise. It does not — #369 had already moved that arm for that reason — and the interpretation reached the user mid-decision.
- **Sentiment written as specification.** A user decision relayed as "burn everything down and make it right" became "`make parity` diverges permanently" in #369's body. The decision was to split the affected model by subcommand; all three parity arms pass.

## Root cause

Both roles reason about the same objects — issues, PRs, files — through different projections. The epic sees a dependency graph; the conductor sees branches, files, and live workers. Each projection looks complete from inside it, and neither role can tell which facts it is missing.

The third error has a distinct and more expensive cause: the epic owns the durable artifact but sits furthest from the freshest information. User decisions land in whichever session is talking to the user. When that is the conductor, the epic learns them secondhand and writes them into the one artifact meant to be citable. A wrong conductor message costs a correction; a wrong EPIC body is read later as settled.

The split is a context-management device — one session cannot hold 9 slots plus 10 EPICs plus a 65-issue backlog, which is why `ba06e87` exists. The roles are not naturally separable, so the fix is to reduce seam traffic and give the remainder a durable record, not to redraw the boundary.

### 1. Create the decision-relay channel, and mark every relay

No skill currently documents a conductor→epic path for user decisions. The only documented pushes in that direction are the `needs-human`/`completed` completion pushes (`bip-conductor/SKILL.md:82-97`, Step 7 at `:272`); everything else in Conventions is epic→conductor→worker. So this item *creates* the convention rather than annotating existing traffic.

Add `### Decision relays: PROVISIONAL and FINAL` to `bip-conductor/SKILL.md` after "Completion pushes": when the conductor is the session talking to the user, it pushes the resulting decision to the `.epic-session` address, prefixed `PROVISIONAL` or `FINAL`. Nothing goes unmarked. Scope is one-directional — epic→conductor relays are unaffected, since the conductor writes no citable artifact.

Mirror the rule into `bip-conductor-poll/SKILL.md:173` (Conventions), following the existing pattern where `:104` duplicates the durable/transient test near-verbatim: a conductor mid-cycle is reading the poll skill, not the cold-start skill.

### 2. The EPIC body records only FINAL shapes

`bip-epic/SKILL.md:140` states "fleet state is derived, never authored." Extend that paragraph: a user decision may be written into an EPIC body only after the deciding session confirms `FINAL` shape. Until then it is provisional and stays out.

### 3. Split Step 4 into two named analyses

`bip-epic/SKILL.md:119` bundles two analyses that share no method: reading issue bodies against each other for dependency direction, and building a file-overlap matrix. Doing the first feels like doing both — that is how "No collisions" got written about two colliding files.

Split into Step 4a (dependency direction) and Step 4b (file overlap), each with its own completion criterion. Step 4b must state that a pair cleared on dependency grounds is not thereby cleared on file grounds.

### 4. Handshake before dispatching Group A

`bip-epic/SKILL.md:57` dispatches scanners with no instruction to consult the conductor first. In the observed run the epic sent its intro and both scanner groups in one message; Group A then spent ~58k tokens re-verifying the state of 10 issues, 9 of which the conductor's own report covered 90 seconds later.

Add a gate with three parts the current draft of this idea lacked:

- **Payload**: the handshake returns the slot→issue occupancy table from `bip-conductor` Step 5, plus the negative list from item 6.
- **Scope**: "what the conductor does not already own" means issues with no live slot.
- **Fallback**: if no conductor session is addressable, proceed with the full Group A dispatch. `bip-epic/SKILL.md:10` states the roles "do not need to be the same session," and Step 6 at `:168` already handles the conductor-absent case; without a matching fallback a solo `/bip-epic` run deadlocks at Step 2.

Group A is not redundant — in the observed run it independently found an open PR and three dependency corrections the conductor lacked — so this narrows it rather than removing it.

This moves a stated boundary, so two claims need softening in the same edit: `bip-epic/SKILL.md:13` ("this skill does not scan clones or tmux, and has no visibility into them") and `:117` ("Clone/tmux occupancy is not part of this reconciliation"). The epic still does not *scan*; it now *consumes* a conductor-supplied table.

### 5. Forward worker findings verbatim and attributed

Worker semantic findings reach the epic through the conductor, which invites the conductor to interpret them. Add `### Forwarding worker findings` to `bip-conductor/SKILL.md` beside item 1's subsection: the worker's finding forwards verbatim and attributed, with any conductor reading marked separately as the conductor's own.

### 6. Record decisions and the negative list to a durable fleet file

The **negative list** — decisions already taken *against* an action, with reasons — is the item the receiving session structurally cannot re-derive: it is absence of work, which leaves no trace in `git`, `gh`, or `tmux`. In the observed run the conductor volunteered such a block unprompted (`SEQUENCING ALREADY APPLIED — do not re-propose these as ready`), and it prevented the epic from proposing `phyz`#2080, whose stated prerequisites had both merged that day so it read as unblocked, and which the conductor had already stood a clone down for.

Items 1 and 5 and this list are all message-carried, which `CONSTITUTION.md` Article II forbids: *"All state that matters MUST survive context compaction and session boundaries. Use files … A fresh agent reading the files MUST be able to pick up where the last one left off."* A `FINAL` marker is what authorizes an EPIC-body write; if the epic compacts between receiving the relay and writing the body, the authorization is gone and item 2 silently reverts to the failure it exists to prevent.

The repo already has the counter-pattern one screen above where item 1 lands — `bip-conductor/SKILL.md:70` requires that a nudge changing what a worker produces "MUST be recorded durably in the same step as the message, by the conductor appending a timestamped, attributed entry."

So: every `FINAL` relay, every forwarded worker finding, and every negative-list entry gets a timestamped, attributed append to `.epic-decisions.md` in the conductor cwd, alongside `.epic-notifications.log`, gitignored the same way as `.epic-status.json` (`bip-conductor/SKILL.md:304`).

**Not `.epic-worklog.md`.** That file is per-slot and is `rm -f`'d on slot cleanup (`bip-conductor-spawn/SKILL.md:140`, `bip-conductor-poll/SKILL.md:152`), so a fleet-level decision written there is destroyed by unrelated housekeeping.

This also reconciles with `bip-conductor-poll/SKILL.md:160-171` ("Filter and route fleet-level findings"), which currently governs this content and routes it elsewhere. The negative list passes that filter's gate 1 (not derived, so not "record nothing"), but its stated destination — "`/bip-epic`'s own memory" — is unavailable wherever the auto-memory directory is deliberately skipped, which `bip-epic/SKILL.md:51` explicitly allows and `phyz` does. `.epic-decisions.md` becomes the destination; "Output structure" (`:109`) reports *from* that file rather than being its home.

## Files to modify

- `skills/bip-epic/SKILL.md` — `:57` Step 2 handshake gate; `:119` Step 4 split; `:140` FINAL-only body rule; `:13` and `:117` boundary softening
- `skills/bip-conductor/SKILL.md` — new `### Decision relays` and `### Forwarding worker findings` subsections after "Completion pushes"; `:221` Step 5 negative list; `:304` gitignore entry for `.epic-decisions.md`
- `skills/bip-conductor-poll/SKILL.md` — `:109` Output structure reports from `.epic-decisions.md`; `:160-171` routing filter reconciliation; `:173` Conventions mirror of the relay-marking rule

## Success criteria

1. `bip-conductor` documents a conductor→epic decision-relay channel requiring a `PROVISIONAL`/`FINAL` marker, and `bip-conductor-poll`'s Conventions mirrors it.
2. `bip-epic/SKILL.md:140` forbids writing a user decision into an EPIC body before `FINAL` confirmation.
3. `bip-epic`'s Step 4 is two numbered sub-steps with separate completion criteria, and 4b states that dependency independence does not imply file independence.
4. `bip-epic`'s Step 2 gates Group A on the conductor handshake, specifies the payload and the "no live slot" scope, and gives a no-conductor fallback that does not deadlock.
5. `bip-conductor` requires verbatim attribution on forwarded worker findings.
6. `.epic-decisions.md` is specified in the conductor cwd, gitignored, and named as the append target for `FINAL` relays, forwarded findings, and negative-list entries; no skill routes fleet-level decisions to the per-slot `.epic-worklog.md`.

## Scope boundaries

**In scope**: the six edits above, to the three named skill files.

**Out of scope**:

- Redrawing which role owns what. The domains are working; only the interface leaks.
- Merging the two skills. The split is load-bearing at 9 slots. A guideline on when a small fleet does not warrant two sessions needs more than one session's evidence.
- **A regression gate for skill structure.** `.github/workflows/ci.yml` is Go-only and nothing in `Makefile` or `scripts/` touches skill markdown, so someone collapsing Step 4a/4b back into one step trips nothing. A `scripts/check-skill-headings.sh` asserting required anchors would be ~15 lines and CI-wireable, but it is new infrastructure for the whole `skills/` tree and belongs in its own issue. Criterion 6 is the one item here that becomes mechanically checkable for free — after a real EPIC run, `.epic-decisions.md` either has timestamped attributed entries or it does not.
- The `needs-human` notification path. `e60ab7c` removed the already-dead bell/`ntfy` snippets in favor of agent-to-agent `SendMessage`; nothing has paged a human for some time. In the observed run the decision still reached the user through the conductor's own conversation, so the gap is narrower than "no channel to the human" — `needs-human` assumes an attended conductor. Worth its own issue if it recurs unattended.

## References

- `ba06e87` — the split this hardens
- `e60ab7c` — completion pushes; its self-registered-address design is the model for item 1
- `0959203` — rescoped Group B as a recency feed, the same class of "this scan does not cover what you think" correction as item 4
- #164 — capability-gated dynamic-workflow path for the fan-out skills; adjacent but disjoint, as it explicitly excludes `bip-epic*` and `bip-spawn`
- `matsengrp/phyz`#369 — the EPIC run that produced the evidence

matsen:
## Summary

`ba06e87` split `bip-epic` (topic strategy) from `bip-conductor` (fleet ops). One 9-slot EPIC run of that split (`matsengrp/phyz`#369) produced three seam errors, all in the interface between the two roles rather than inside either one's own domain: a missed file collision (`bip-epic` cleared two issues as dependency-independent while both edited the same script), a worker finding forwarded by the conductor with its own reading blended into the worker's, and user sentiment relayed through the conductor written into an EPIC body as if it were the decision's substance.

Six edits, documentation-only (`skills/bip-epic/SKILL.md`, `skills/bip-conductor/SKILL.md`, `skills/bip-conductor-poll/SKILL.md`):

1. **Decision-relay channel, marked `PROVISIONAL`/`FINAL`.** New `### Decision relays` subsection in `bip-conductor` — when the conductor is the session talking to the user, it pushes the resulting decision to `/bip-epic`'s self-registered address, prefixed one or the other. Mirrored in `bip-conductor-poll`'s Conventions (a conductor mid-cycle reads that skill, not the cold-start one).
2. **`FINAL`-only EPIC body writes.** `bip-epic`'s "Fleet state is derived" section gains a rule: a user decision may be written into an EPIC body only against a `FINAL` relay (or a decision reached directly in that session) — never a paraphrase of sentiment from a `PROVISIONAL` one.
3. **Step 4a/4b split.** `bip-epic` Step 4 (dependency-direction + file-overlap) had bundled two analyses that share no method. Split into 4a (dependency direction) and 4b (file overlap), each with its own completion criterion; 4b states explicitly that clearing a pair on dependency grounds does not clear it on file grounds — the exact gap that produced the missed collision.
4. **Conductor handshake before Group A.** `bip-epic` Step 2 now resolves the conductor's address and requests its slot→issue occupancy table (issue + **phase**) before fanning out Group A subagents, and scopes each subagent to skip its redundant per-issue "confirm still open" check only for slots in an **in-progress** phase (`coding`/`exploring`/`testing`/`awaiting-results`/`quality-gate`). `completed`/`needs-human` slots are always re-confirmed, since a slot can hold an already-closed issue right up until the separate slot-cleanup housekeeping step runs — occupancy is evidence about the fleet, not about GitHub. Falls back to the full unscoped dispatch when no conductor is addressable, so a solo `/bip-epic` run doesn't deadlock. (In the motivating run, this specific re-verification cost ~58k tokens confirming state the conductor's own report covered 90 seconds later — Group A's independent judgment, e.g. catching an open PR or a dependency issue the conductor's table can't see, is unaffected.)
5. **Verbatim, attributed forwarding.** New `### Forwarding worker findings` subsection in `bip-conductor` — a worker's finding forwards as-is, with any conductor reading marked as a separate, clearly attributed line.
6. **`.epic-decisions.md`.** A durable, gitignored, conductor-cwd file (timestamped markdown, append-only, chronological — same convention as `.epic-worklog.md`, distinct file) that every `FINAL` relay, every forwarded worker finding, and every negative-list entry (decisions already taken *against* an action, with reasons — the one category of fleet fact that leaves no trace in `git`/`tmux`/`gh`) appends to. Explicitly not `.epic-worklog.md`, which is per-slot and `rm -f`'d on cleanup. `bip-conductor-poll`'s routing filter and Output structure now point fleet-level decisions here instead of `/bip-epic`'s own memory, which is unavailable wherever a setup deliberately skips the auto-memory directory.

Two existing boundary claims (`bip-epic`'s intro and its Step 3 reconciliation note) are softened to reflect item 4: the epic still does not *scan* clone/tmux state itself, but it now *consumes* a conductor-supplied table when one is offered.

## Out of scope (per the issue)

Redrawing role ownership, merging the two skills, and a CI-checkable regression gate for skill-heading structure are all explicitly deferred — see the issue's Scope boundaries.

## DEFERRED

- `.epic-decisions.md` needs a real `.gitignore` entry in each EPIC-consuming repo (this repo's own `.gitignore` doesn't list the fleet-state files at all, since they're created per-consuming-repo, not here). First conductor session to append will dirty a worktree until that repo's `.gitignore` is updated — a one-line addition per repo, not blocking this PR.

## Test plan

- Documentation-only change to three `skills/*/SKILL.md` files; no Go code touched (`go build ./...` unaffected).
- Read each edited section against the issue's six success criteria and the cross-file references (`bip-epic` ↔ `bip-conductor` ↔ `bip-conductor-poll`) for internal consistency — no dangling "Step 4" references, no stale line-number pointers.
- Reviewed by `phyz-9f` (the epic session from the `phyz`#369 run this issue is based on) against the six success criteria; one HIGH finding (item 4's occupancy-skip didn't account for `completed`/`needs-human` slots holding already-closed issues) fixed in place.

Closes #207

### DIFF
--- a/skills/bip-conductor-poll/SKILL.md
+++ b/skills/bip-conductor-poll/SKILL.md
@@ -115,9 +115,11 @@ Skip the nudge entirely for a worker in `awaiting-results` with a live `check_cm
 
 3. **Needs human**: Clones in `needs-human` phase — show the lead's assessment and what decision is needed.
 
-4. **Recently landed** (brief): PRs merged since last poll, only if noteworthy.
+4. **Decisions and negative list**: read new entries from `.epic-decisions.md` in the conductor cwd since the last poll (`FINAL` relays, forwarded worker findings, negative-list entries — see `/bip-conductor`'s Conventions, "Decision relays" and ".epic-decisions.md: the durable fleet-decision log") and report them here; this is the file's canonical output surface for a poll cycle, not a fresh derivation.
 
-5. **Execute pending spawns**: If spawn intent files and idle clones both exist, propose running `/bip-conductor-spawn` for them.
+5. **Recently landed** (brief): PRs merged since last poll, only if noteworthy.
+
+6. **Execute pending spawns**: If spawn intent files and idle clones both exist, propose running `/bip-conductor-spawn` for them.
    Wait for confirmation.
 
 ### Housekeeping (do silently, don't report unless problems)
@@ -167,13 +169,16 @@ Before recording anything anywhere, run each candidate fleet-level finding throu
 2. **Is it already recorded?**
    A finding that produced an issue, PR, test, or doc needs no second copy.
 
-Only what survives both gates gets a destination: a durable clone-pool layout decision → a `CLAUDE.md`; a workflow rule → a skill; a topic-level decision → `/bip-epic`'s own memory, not here.
+Only what survives both gates gets a destination: a durable clone-pool layout decision → a `CLAUDE.md`; a workflow rule → a skill; a **fleet-level decision** (a `FINAL` relay, a forwarded worker finding, a negative-list entry) → `.epic-decisions.md` in the conductor cwd, not `/bip-epic`'s own memory — that memory is unavailable wherever the auto-memory directory is deliberately skipped (`bip-epic/SKILL.md`'s Step 1 explicitly allows this, and some setups do it), and `.epic-decisions.md` has no such gap since it's a plain file, not a memory-directory feature.
 Nothing else fits → the poll report itself.
 
 ## Conventions
 
 Same as `/bip-epic`: `iN`/`pN` prefixes, full URLs on first mention.
 Tmux windows named `NNN-YYY` where NNN is the issue number and YYY is the clone/slot name (e.g. `281-cedar` in clone mode, `281-issue-281` in worktree mode).
+
+**Decision relays are marked `PROVISIONAL` or `FINAL`, nothing unmarked.**
+When this conductor session is the one talking to the user mid-poll and a decision results, push it to `/bip-epic` via `$CLONE_ROOT/.epic-session` prefixed `PROVISIONAL` or `FINAL` and append it to `.epic-decisions.md` — see `/bip-conductor`'s Conventions, "Decision relays: PROVISIONAL and FINAL", for the full mechanics; this is the same rule, restated here because a conductor mid-cycle is reading this skill, not the cold-start one (the durable/transient nudge test above duplicates the same way).
 
 ## Layout config (issue #149)
 
--- a/skills/bip-conductor/SKILL.md
+++ b/skills/bip-conductor/SKILL.md
@@ -95,6 +95,46 @@ To push, a role reads the other's file for the exact address and `SendMessage`s 
 A failed `SendMessage` to a drifted name comes back with `success: false` and a "Did you mean: ..." suggestion list of other live sessions — **never act on that suggestion**; trying it is exactly the guessing this design exists to avoid, just prompted by the tool instead of self-initiated.
 The residual risk this doesn't close — an unrelated session claiming the exact same name string in the gap before a refresh — is accepted as rare; the push is a latency optimization, never a hard dependency.
 
+### Decision relays: PROVISIONAL and FINAL
+
+No skill documents a conductor→epic path for user decisions before this — the only documented pushes in that direction are the `needs-human`/`completed` completion pushes above; everything else in these Conventions runs epic→conductor→worker.
+This section creates that convention rather than annotating an existing one.
+
+When the conductor is the session talking to the user and a decision results, push it to `/bip-epic` rather than letting it sit only in this session's own conversation: read `$CLONE_ROOT/.epic-session` for the epic's self-registered address (same mechanics as "Completion pushes" above) and `SendMessage` it the decision, **prefixed `PROVISIONAL` or `FINAL`. Nothing goes unmarked.**
+
+- **`PROVISIONAL`**: the decision is still being discussed, or the conductor is relaying a first read before the user has confirmed it. `/bip-epic` may note that a decision is pending but must not write it into an EPIC body — see that skill's "Fleet state is derived" section for the FINAL-only body rule this feeds.
+- **`FINAL`**: the user has confirmed the decision's shape. This, and only this, authorizes `/bip-epic` to write the decision into an EPIC body.
+- Append every `FINAL` relay (and, once it firms up, the `PROVISIONAL` relay it followed from) to `.epic-decisions.md` in the conductor cwd, timestamped and attributed — see ".epic-decisions.md: the durable fleet-decision log" below. A relay that lives only in the message is gone the moment either session compacts; the body-write authorization in `/bip-epic` depends on being able to re-derive that a `FINAL` marker was actually sent.
+
+This is one-directional: epic→conductor relays (spawn intent, worker corrections) are unaffected by this rule, since the conductor writes no citable artifact from them.
+
+### Forwarding worker findings
+
+A worker's semantic finding — "this assumption doesn't hold," "this arm was already moved" — reaches the epic through the conductor, and passing it through invites the conductor to interpret it along the way.
+Don't. Forward the finding **verbatim and attributed** to the worker (issue/slot) that produced it.
+If the conductor has its own reading of the finding worth adding, add it as a separate, clearly marked line — never blended into the forwarded text so that the epic (or the user) cannot tell which parts are the worker's observation and which are the conductor's interpretation.
+The failure this prevents: a worker's matrix-provenance finding, forwarded as if it undercut a sibling issue's premise, when the EPIC had already moved that arm for the same reason — the conductor's reading reached the user mid-decision as though it were the worker's own conclusion.
+Append forwarded findings to `.epic-decisions.md` alongside decision relays (see ".epic-decisions.md: the durable fleet-decision log" above), same reasoning: message-only state does not survive compaction.
+
+### `.epic-decisions.md`: the durable fleet-decision log
+
+Every `FINAL` relay (above), every forwarded worker finding (above), and every negative-list entry (Step 5's dashboard) appends here — in the conductor cwd, gitignored the same way as `.epic-status.json` (see the gitignore note near that spec below).
+Format is timestamped markdown, append-only, never edit a previous entry — the same convention as `.epic-worklog.md`, because this is attributed narrative for a human (or a fresh epic session) to read, not a machine-parsed event stream like the JSONL `.epic-notifications.log`.
+
+```
+## 2026-08-28T14:30:00Z — RELAY:FINAL (conductor→epic)
+<the decision, restated in full, not a paraphrase of sentiment>
+
+## 2026-08-28T14:32:00Z — FINDING (worker i2098, forwarded by conductor)
+<the worker's finding, verbatim>
+Conductor's own reading, if any, marked separately: <...>
+
+## 2026-08-28T09:00:00Z — NEGATIVE (conductor)
+<action not taken, and why — e.g. "i2080 not proposed: clone stood down 08-27 for <reason>, prerequisites merging today doesn't reopen it">
+```
+
+**Not `.epic-worklog.md`.** That file is per-slot and gets `rm -f`'d on slot cleanup, so a fleet-level entry written there is destroyed by unrelated housekeeping the moment that slot is cleaned up.
+
 ## Configuration
 
 The conductor skill reads `.epic-config.json` from the repo root — the same file `/bip-epic` reads.
@@ -231,6 +271,11 @@ The dashboard is **slot-centric** — the epic's dashboard is issue-centric; thi
 
 **Pending spawn intent**: list any `.spawn-prompts/` files found in Step 3/4 (either naming pattern), with the available slots that could run them.
 
+**Negative list**: also surface decisions already taken *against* an action, with reasons — sequencing already applied, an issue already stood down for a reason that would otherwise look resolved, and similar.
+This is the one category of fleet fact the epic (or a fresh conductor) cannot re-derive from `git`/`tmux`/`gh`: it's the absence of work, which leaves no trace in any of those.
+Read `.epic-decisions.md` in the conductor cwd (see Conventions, "Decision relays" and ".epic-decisions.md: the durable fleet-decision log") for prior entries and append any new one here, timestamped and attributed, in the same step you surface it — don't let it live only in this dashboard render.
+Concrete shape from the run that motivated this: an issue whose stated prerequisites both merged the same day reads as unblocked, but the conductor had already stood a clone down for it for an unrelated reason — without the negative list, that reads as ready to the epic and gets proposed again.
+
 ### Step 6: Propose next action
 
 First, do housekeeping automatically (no need to ask):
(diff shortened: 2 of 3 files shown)


matsen:
Review from the author of #207 and the epic session whose run produced its evidence.

**Five of six criteria pass cleanly.** SC1 (relay channel + poll mirror), SC2 (FINAL-only body rule, with the "burn everything down" paraphrase example carried into the skill text — good, that's the failure verbatim), SC3 (4a/4b with separate completion criteria and the explicit "cleared in 4a is not thereby cleared here"), SC5 (verbatim forwarding), SC6 (`.epic-decisions.md` with worked example, format rationale, the `.epic-worklog.md` exclusion, gitignore line, and the poll routing-filter reconciliation). The markdown-over-JSONL call is right and the rationale given for it is the correct one.

## SC4 — one HIGH: the scope rule disables the check most likely to fire

> **Scope**: ... have it skip step 3's `gh issue view` confirmation for any child item the table already shows occupied — the conductor's own dashboard just reconfirmed that item is open and active.

Occupancy does not imply open. `completed` is a live phase (`bip-conductor/SKILL.md:313`), and slot cleanup is a *separate* housekeeping step (`bip-conductor-poll/SKILL.md:152`), so there is a standing window where a slot is occupied by a finished worker whose issue is already closed.

That window is not hypothetical — it is where the motivating run actually sat. At the moment of that Group A dispatch, the conductor's report listed `spruce` holding phyz#2086 and Group A independently found #2086 **closed at 22:26**, which is what exposed the EPIC body's "Ready now: #2097, #2087, #2086" line as stale. Under this rule as written, #2086's slot reads occupied, the `gh issue view` is skipped, and that stale line survives the sweep.

So the optimization removes the check precisely in the case it most needs to run: a slot that just finished is *more* likely to have a closed issue than an idle one, not less.

Suggested fix, narrow — gate on phase, not on occupancy. Have the handshake payload carry each slot's phase alongside its issue, and skip the confirmation only for phases that imply work still in progress (`coding`, `exploring`, `testing`, `awaiting-results`, `quality-gate`). Never skip for `completed` or `needs-human`. That keeps essentially all of the ~58k-token saving — most slots at any moment are mid-flight — while leaving the terminal phases, which are both rarer and the only ones that can be stale, fully checked.

Worth stating in the skill text too: an occupied slot is evidence about *the fleet*, not about *GitHub*.

## Two minor

**Worked example is not chronological.** The `.epic-decisions.md` sample runs `14:30:00Z`, `14:32:00Z`, then `09:00:00Z`. For a file the same block calls "append-only, never edit a previous entry," a reader copying the shape learns the wrong thing. Reorder so the NEGATIVE entry is first or restamp it later.

**`.epic-decisions.md` needs a real gitignore entry, outside this repo.** `bip-conductor/SKILL.md:346` now says it must be gitignored, but the file lands in the conductor cwd — a consuming repo. phyz's `.gitignore:168-169` currently lists `.epic-status.json` and `.epic-worklog.md` and not the new file, so it will show up untracked in every clone until someone adds it. Not fixable here; worth a line in the PR body or a follow-up issue against consuming repos, so the first conductor to append doesn't dirty a worktree.

## On the Step 2 scoping call

Splitting "skip the redundant per-item open/closed query" from "skip whole subagents" is the right read of item 4 and better than what #207 specified — occupancy and dependency reasoning genuinely are different questions, and that same run is the evidence (Group A found an open PR and three dependency corrections the conductor's table had no way to see). Keep that framing; only the phase gate above needs adding.

Item 6 remains the reasoned-not-witnessed one. Flagging it as such in the PR body was the right call.

matsen:
SC4 re-checked at 611254e — passes. All six criteria now met.

The phase gate is right, and it's encoded consistently in both places it has to be: the Step 2 `Scope` bullet and the Group A subagent brief at `:91` carry the same phase list and the same rationale, so a subagent reading only its own brief gets the correct rule. That consistency was the thing most likely to go wrong in this fix, since the earlier version had encoded the occupancy rule in both spots.

"Occupancy is evidence about the fleet, not about GitHub" made it into the skill text, and the reason a finished slot is *more* dangerous than a mid-flight one — finishing is what triggers the issue closing, and cleanup lags — is stated rather than left for the reader to reconstruct. That's the part that keeps someone from re-optimizing this back in six months.

Timestamps now chronological (09:00 → 14:30 → 14:32). The consuming-repo `.gitignore` gap is correctly DEFERRED rather than fixed here — it's a one-line addition per consuming repo and genuinely can't land in this one.

No further findings from me. Ready to merge as far as this review is concerned.

One note for whoever runs the next EPIC: item 6 (`.epic-decisions.md`) is the only part of this that's reasoned rather than witnessed — it came from the pre-file review catching a Constitution II violation, not from an observed failure. Worth watching on the first real run whether appending to it every cycle is ergonomic for the conductor, or whether it decays into a write-only log nobody reads. The negative-list entries are the ones with demonstrated value; the `FINAL` relay entries are the speculative part.

PR #210, opened 2026-08-29
Require citation resolution before acting in bip-epic/bip-conductor

### LINKED ISSUE #209: Make citation resolution explicit in the cross-session verification loop
# Make citation resolution explicit in the cross-session verification loop

🤖

`bip-epic`'s Step 4b and `bip-conductor`'s fleet-facts conventions gain one rule: **resolve a citation before acting on it, and never fill in a missing path component from surrounding context.** Plus a smaller companion rule for corrections that reduce a reported number to nothing.

## Motivation

#207 (PR #208) made two sessions verify each other's claims at the seam. That worked — it caught a file collision, a wrong premise about a matched-model arm, and a false parity claim, all within minutes. But one full session of running it (EPIC `matsengrp/phyz`#369, 2026-08-28) produced **four failures of the verification step itself**, all the same shape: the check executed correctly, against the wrong resolution context, and returned a confident wrong answer.

The failure is silent, and often *inverted* — it makes correct citations read as broken:

1. **Inherited cwd.** A `cd` in one `Bash` call persisted into later calls. Three real files read as "No such file or directory," and the session was one step from "correcting" an issue draft whose citations were already right.
2. **Reconstructed directory.** A citation arrived as `run_heavy_baselines.py:30-43` with no directory. The receiving session supplied one from surrounding context, ran `find` against its own invention, got a real "missing" result, and reported it as the sender's error. **The `find` made it more convincing, not less.**
3. **Same name, different file.** Three experiments define `ARM_B_CLUSTERS` — at `:73`, `:59` and `:41` of three separate `scripts/common.py` copies. Two sessions cited different ones and each read the other as wrong. (They turned out to define the same set by different decompositions, which took another round to establish.)
4. **Same name, different object.** `ECMK07` names two different matrices — one PRANK-derived, one Kosiol — differing by up to 175 nat on fixed trees. A whole thread of reasoning ran before anyone checked they were the same table.

Two of these cost real work: a spawn brief shipped saying "No collisions" about two issues editing the same file, and a correction that would have un-fixed a correct issue had it been applied.

## Root cause

In a repo with per-experiment `scripts/` copies, a bare filename plus line number is not an address. Neither is a path resolved against an unstated working directory, nor a symbol name shared across modules. But every one of these *looks* like an address, and the tools confirm whatever they are handed. Verification tells you a claim is consistent with the context you resolved it in; it does not tell you the context was right.

This matters more under #207 than it did before, because cross-session citation checking is now routine rather than incidental.

### 1. Resolve before acting (`bip-epic` Step 4b, `bip-conductor` conventions)

**`skills/bip-epic/SKILL.md`, Step 4b** — file-overlap detection extracts paths from issue bodies, which is exactly where this bites. Add: a path extracted from prose is a *candidate*, not a location. Resolve it with `find`/`grep -r` before recording an overlap or its absence. A filename with no directory component is unresolved; two issues naming `common.py` are not evidence of overlap.

**`skills/bip-conductor/SKILL.md`, near "Forwarding worker findings"** — add: never reconstruct a missing path component from surrounding context. Resolve it, or ask the sender. A reconstructed path that then fails a `find` is not a verified error; it is an invented one with a real-looking result attached.

**Both** — verification shell calls should carry their own working directory (`cd` inside the same command, or absolute paths) rather than relying on inherited cwd.

### 2. A correction that reduces a number to nothing deserves a second look

Same loop, opposite direction, one instance rather than four — but the fix is one sentence and the failure reached a citable artifact.

Having found that a reported sign flip was guaranteed by construction (the new target maxed over a strict subset of the old one's candidates, so every pair was certain to move that way), a session concluded the reported flip *count* therefore "carries no information." It does: crossing zero requires the move to exceed the original magnitude, which the direction argument does not guarantee. The count was real evidence about the contamination's size. The same number had been recorded as a finding that morning and was about to be recorded as nothing by evening; neither was right.

**Add to `bip-epic` Step 5** (EPIC body update): when a correction would reduce a previously-reported quantity to *no* information, check whether a weaker claim survives before writing the stronger retraction. Debunking has its own momentum, and "carries no information" is a more quotable sentence than "carries less than it appears to, in this specific way."

## Files to modify

- `skills/bip-epic/SKILL.md` — Step 4b resolution rule; Step 5 over-correction check
- `skills/bip-conductor/SKILL.md` — no-reconstruction rule beside "Forwarding worker findings"

## Success criteria

1. `bip-epic` Step 4b states that an extracted path is a candidate until resolved, and that a filename without a directory component is unresolved.
2. `bip-conductor` forbids reconstructing a missing path component from context, and names the specific trap: a `find` run against a reconstructed path returns a real result for an invented question.
3. Both skills say verification commands should carry their own working directory rather than inherit one.
4. `bip-epic` Step 5 requires checking for a surviving weaker claim before recording a total retraction.

## Scope boundaries

**In scope**: the two skill files above.

**Out of scope**:

- **The phyz-side half.** That three experiments define `ARM_B_CLUSTERS` through three expressions, and that `ECMK07` names two matrices, are facts about `matsengrp/phyz` and belong in its `experiments/CLAUDE.md` — a separate decision in a separate repo, not fixable here.
- **Tooling to detect stale cwd or ambiguous symbols.** A linter for skill-adjacent shell usage is a much larger piece of work than the rule it would enforce, and the rule is worth having first.
- **Rewriting #207's verification loop.** It works. This hardens one step inside it.

## References

- #207 / PR #208 — the seam hardening that made cross-session verification routine
- `3930354` — the gitignore note generalization, which came from the same session and the same "the note as written fixes one instance and leaves the class" observation
- `matsengrp/phyz`#369 — the EPIC run that produced all four instances

matsen:
## Summary

One full EPIC run (`matsengrp/phyz`#369, under #207/PR #208's new cross-session verification loop) produced four failures of the verification step itself, all the same shape: a check executed correctly against the wrong resolution context and returned a confident wrong answer (inherited `cd`, a path reconstructed from surrounding context, a symbol/filename shared across per-experiment copies). A fifth, related failure: a correction that found a reported number's *direction* was guaranteed by construction also concluded — wrongly — that the number "carries no information."

- `skills/bip-epic/SKILL.md` Step 4b: a path extracted from prose is a candidate, not a location — resolve with `find`/`grep -r` before recording an overlap or its absence; a filename with no directory component is unresolved.
- `skills/bip-conductor/SKILL.md`, new "Resolving a citation before acting on it" section beside "Forwarding worker findings": never reconstruct a missing path component from context — resolve it or ask the sender. Names the trap directly: a `find` against a reconstructed path returns a real result for an invented question.
- Both skills: verification shell calls should carry their own working directory (`cd` in the same command, or absolute paths) rather than relying on inherited cwd.
- `skills/bip-epic/SKILL.md` Step 5: a correction that would reduce a previously-reported quantity to *no* information should first check whether a weaker, still-true claim survives.

## Test plan

- Documentation-only change to two skill files; no code paths affected.
- Read both edited sections in context to confirm they fit the surrounding structure and don't duplicate existing rules.

Closes #209

### DIFF
--- a/skills/bip-conductor-poll/SKILL.md
+++ b/skills/bip-conductor-poll/SKILL.md
@@ -180,6 +180,9 @@ Tmux windows named `NNN-YYY` where NNN is the issue number and YYY is the clone/
 **Decision relays are marked `PROVISIONAL` or `FINAL`, nothing unmarked.**
 When this conductor session is the one talking to the user mid-poll and a decision results, push it to `/bip-epic` via `$CLONE_ROOT/.epic-session` prefixed `PROVISIONAL` or `FINAL` and append it to `.epic-decisions.md` — see `/bip-conductor`'s Conventions, "Decision relays: PROVISIONAL and FINAL", for the full mechanics; this is the same rule, restated here because a conductor mid-cycle is reading this skill, not the cold-start one (the durable/transient nudge test above duplicates the same way).
 
+**Resolve a citation before acting on it, here too.**
+A worker's finding forwarded mid-poll carries the same citation risk as at cold start — a bare filename with no directory is unresolved, and reconstructing one from context produces a real `find` result for an invented question. See `/bip-conductor`'s Conventions, "Resolving a citation before acting on it", for the full mechanics; restated here for the same reason as the decision-relay rule above.
+
 ## Layout config (issue #149)
 
 `.epic-config.json` keeps working.
--- a/skills/bip-conductor/SKILL.md
+++ b/skills/bip-conductor/SKILL.md
@@ -116,6 +116,13 @@ If the conductor has its own reading of the finding worth adding, add it as a se
 The failure this prevents: a worker's matrix-provenance finding, forwarded as if it undercut a sibling issue's premise, when the EPIC had already moved that arm for the same reason — the conductor's reading reached the user mid-decision as though it were the worker's own conclusion.
 Append forwarded findings to `.epic-decisions.md` alongside decision relays (see ".epic-decisions.md: the durable fleet-decision log" above), same reasoning: message-only state does not survive compaction.
 
+### Resolving a citation before acting on it
+
+A finding often carries a citation — a file, a line range, a symbol — and forwarding it verbatim (above) does not excuse skipping resolution before acting on it locally. Never reconstruct a missing path component from surrounding context: a citation like `run_heavy_baselines.py:30-43` with no directory is unresolved, not incomplete-but-inferable. Resolve it against the repo (`find`/`grep -r`) or ask the sender which copy they meant.
+**A `find` run against a reconstructed path returns a real result for an invented question.** The result looking concrete does not make the resolution correct, and a "missing file" found this way is not evidence of an error in the citation — it is evidence of an error in the reconstruction.
+The same caution applies to a bare symbol name in a repo with duplicated modules: two files can define the same name for different things, and matching on the name alone is not resolution.
+Verification shell calls should carry their own working directory — `cd` inside the same command, or an absolute path — rather than relying on a `cd` from an earlier call in the same session persisting into this one.
+
 ### `.epic-decisions.md`: the durable fleet-decision log
 
 Every `FINAL` relay (above), every forwarded worker finding (above), and every negative-list entry (Step 5's dashboard) appends here — in the conductor cwd, gitignored the same way as `.epic-status.json` (see the gitignore note near that spec below).
--- a/skills/bip-conductor/SKILL.md
+++ b/skills/bip-conductor/SKILL.md
@@ -122,6 +122,7 @@ A finding often carries a citation — a file, a line range, a symbol — and fo
 **A `find` run against a reconstructed path returns a real result for an invented question.** The result looking concrete does not make the resolution correct, and a "missing file" found this way is not evidence of an error in the citation — it is evidence of an error in the reconstruction.
 The same caution applies to a bare symbol name in a repo with duplicated modules: two files can define the same name for different things, and matching on the name alone is not resolution.
 Verification shell calls should carry their own working directory — `cd` inside the same command, or an absolute path — rather than relying on a `cd` from an earlier call in the same session persisting into this one.
+This cuts both ways: when relaying a finding or citing a file in a message to another session — a worker report, a push to the epic — include the directory. A bare filename plus line number is not an address, and a sender who includes the directory removes the ambiguity at its cheapest point, before it costs the receiver a resolution or a round-trip question.
 
 ### `.epic-decisions.md`: the durable fleet-decision log
 
--- a/skills/bip-epic/SKILL.md
+++ b/skills/bip-epic/SKILL.md
@@ -147,16 +147,16 @@ Run this before proposing any issue as ready to spawn, not after — once a bran
 
 This is a distinct analysis from 4a, not a second pass over the same reading — it automates what has so far been a manual capability: extracting file paths from issue bodies and building an overlap matrix.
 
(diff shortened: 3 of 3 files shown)


matsen:
Review from #209's author and one of the two sessions whose failures produced it.

**All four success criteria met.** SC1 (path-is-a-candidate, filename-without-directory-is-unresolved), SC2 (no reconstruction, with the `find`-against-a-reconstructed-path trap named outright), SC3 (working directory in both skills), SC4 (weaker-claim check in Step 5). The Step 5 generalization — "guaranteed in direction by construction is not thereby guaranteed in magnitude" — is the right abstraction of that failure rather than a restatement of the instance.

Verified separately, since this PR is about exactly this class of problem: **no skill or doc in the repo cites a line number into `bip-epic`, `bip-conductor`, `bip-conductor-poll` or `bip-conductor-spawn`.** All cross-references are by section name, so the 16 added lines break nothing. Worth knowing that this repo is already immune to the shift-the-line-numbers problem by convention.

## One substantive gap: the rule is receiver-side only

Both new passages harden the *receiver*: resolve it, don't reconstruct it, ask the sender which copy they meant. Nothing addresses the *sender*.

That matters because the sender is where the failure was cheapest to prevent, and the sender in the motivating instance was me. My message carried `run_heavy_baselines.py:30-43` with no directory. The conductor then reconstructed a directory from context and reported a nonexistent error. The receiver-side rule correctly stops the second half — but if senders keep emitting bare filenames, every one now costs either a resolution or a round-trip question, in a loop where round trips are the expensive part.

Suggested addition, one sentence, either skill or both:

> When citing a file in a message to another session, include the directory — in a repo with per-experiment `scripts/` copies, a bare filename plus line number is not an address.

That is strictly cheaper than resolution and removes the ambiguity at source rather than catching it downstream.

## Two minor

**Step 4b placement.** The new paragraph sits between numbered item 4 and the two **Completion criterion** lines, which separates the procedure from its own criterion. The rule is really about item 1 (where paths get extracted from prose), so a reader working 1→4 meets it only after doing the extraction it governs. Consider moving it above the numbered list as a preamble, or folding the first sentence into item 1 and leaving the elaboration where it is.

**`bip-conductor-poll` mirror not done.** #207/PR #208 established the pattern of duplicating load-bearing conductor conventions into the poll skill, on the stated grounds that "a conductor mid-cycle is reading this skill, not the cold-start one" (`bip-conductor-poll:104` mirrors the durable/transient nudge test that way, and #208 added the relay-marking mirror). A conductor resolving a citation mid-poll is in the same position. #209's Files-to-modify named only two files so this PR is correct as scoped — flagging it as a question for you rather than a defect: is the mirror wanted here, or is citation resolution rare enough mid-poll that the cold-start statement suffices?

Nothing blocking. The sender-side sentence is the one I'd want before merge; the other two are yours to take or leave.

matsen:
🤖 Addressed the two concrete points:

- **Sender-side rule added** to both skills: "include the directory when citing a file in a message to another session" — one sentence in `bip-conductor`'s new section and one in `bip-epic` Step 4b, each noting a bare filename+line is not an address and that the sender including it is cheaper than a receiver-side resolution or round trip.
- **Step 4b placement fixed**: folded "a path extracted this way is a candidate, not a location" into item 1, where the extraction actually happens, and left the `find`/`grep -r` elaboration after the list.

The `bip-conductor-poll` mirror question is open — asking the user directly since #209 scoped only the two files and this would expand that.

matsen:
🤖 Added the `bip-conductor-poll` mirror (user opted to add it now rather than skip or defer): a short restatement of the resolve-before-acting rule pointing back to `bip-conductor`'s Conventions section, following the same pattern PR #208 used for the decision-relay rule there.

PR #214, opened 2026-09-03
Catch renaming: enforcement hooks and skill passes

### LINKED ISSUE #213: Catch renaming: terminology-checker agent and enforcement hooks
> Claude Code:

Catch a thing being given a second name, with something that reads the text rather than relying on the reader to notice.

The rule already exists in prose — "Once a thing has a name, that name is fixed" — and prose alone does not hold it. In a long session an agent will call something `requeues`, then `rescue` a message later; will write `crossings` for what the user has been calling shared molecules; will call one quantity `detection response` in one place and `depth response` in another. The reader cannot tell what the new word means and has to guess.

The failure is narrow and worth stating exactly: it is not inventing a name for something new, which is fine and sometimes necessary. It is replacing a name that already exists, usually one word for one word, with nothing gained.

## Where it happens

Prose leaves an agent roughly ten thousand times a session, across chat, `Write`/`Edit`, and Bash. Bodies posted to GitHub are about 3% of that, and the least likely 3%: such a body is written in one sitting with the whole thing in view, while renaming is a drift across hours. Counted across the transcripts on one machine, all three cases above happened in chat and none reached a post.

That rules out putting the check on GitHub posts, which was the first plan here. It has to watch the chat, and be cheap enough to run every turn.

## What gets added

- `hooks/novel-words.py` — a `Stop` hook reporting words in the turn just finished that nobody has used earlier in the session. A substituted name is by construction new and the name it replaces is old, so the candidates need no model.
- `hooks/terminology.sh` — a `UserPromptSubmit` hook restating the rule each turn, so it does not decay as a session grows past where the instructions were read.
- The renaming rule in `bip-remove-metaspeak`, `bip-ms-sweep` and `bip-issue-check`, which already read a document end to end, so it costs no extra pass.
- A `Makefile` target installing hooks alongside agents and skills, and `hooks/README.md` for the `settings.json` entries a user has to add by hand.

## Tried, does not work

Recorded so nobody builds them again.

**A blocklist of banned words.** The substituted word is unknown in advance — that is what makes it a substitution. Nothing can be listed.

**A dictionary or stoplist filter over candidate words.** `rescue`, `crossings`, `residue` and `record` are all ordinary English. Filtering to non-dictionary words removes exactly the real cases. Novelty *to the session* is the signal; novelty to English is not.

**Example words in `terminology.sh`.** An earlier version listed pinned, route, burden, compartment, signature, residue, snap, loose. An instruction that lists words reads as a list of banned words, and `residue` is the correct domain term in one of the projects the hook fires in. Worked examples belong with the skills, which read someone else's text rather than writing their own.

**Stating the rule and nothing more.** It was already in `CLAUDE.md`, correctly worded, and the agent that read it renamed three things the same day — after adding a hook restating it every turn. Instruction without detection does not hold. `terminology.sh` is kept only because it is nearly free.

**A subagent reading documents for renamings.** 79 seconds a run after its reading was capped, from 206 before. Worse, a subagent never receives the conversation, so its first authority — the word the user used — is out of reach and it can only find a document contradicting itself. That is the smaller half of the problem at a price that keeps it off any routine path.

**A hook refusing GitHub posts until the body was checked.** It existed only to force that subagent to run. Its own ledger on the day it was built: four refusals, three of them its own defects, and the two real catches came from the subagent rather than from it.

**Reporting every word new to the session.** Fires on 31.9% of turns over 3,622 turns in three sessions. Requiring the word to recur in a later turn is no better, at 29.7%. A message on a third of turns is one that gets skipped past, which is the failure this exists to avoid. Requiring two uses within one turn gives 4.8%, and ignoring plurals and verb forms of words already seen gives 3.8%.

**Symlinking hooks into this repo,** as `make install` does for agents and skills. The symlink dangles whenever another branch is checked out, the hook command exits 127, and Claude Code treats that as a non-blocking error — the check stops running with nothing said. Hooks are copied instead.

## Known limits of what was built

`novel-words.py` reports candidates and cannot tell whether a new word names something that already has a name; that judgement stays with the agent. Of two known renamings in the measured transcripts it catches one. It watches chat only, so a renaming that first appears in a file or a post is caught by the skills or not at all.

psathyrella:
NOTE: not ready for merging. I'm trying out some different approaches to reduce how frequently it replaces existing jargon with new jargon.

**UPDATE** seems good to go

> Claude Code:

Closes #213.

Adds a `Stop` hook that reports words new to the session, a `UserPromptSubmit` hook restating the rule each turn, and the renaming rule itself in the three skills that already read a document end to end.

The rule exists in prose already — "Once a thing has a name, that name is fixed" — and prose alone does not hold it. An agent will call something `requeues`, then `rescue` a message later; will write `crossings` for what the user has been calling shared molecules; will call one quantity `detection response` in one place and `depth response` in another. The rule is read once at session start, and nothing afterwards reads what was written.

The defect is narrow: replacing a name that already exists, usually one word for one word, with nothing gained. Coining a name for something genuinely new is not the problem.

## Where the prose is

Prose leaves an agent roughly ten thousand times a session, across chat, `Write`/`Edit`, and Bash. Bodies posted to GitHub are about 3% of that, and the least likely 3%: such a body is written in one sitting with the whole thing in view, while renaming is a drift across hours. Counted across the transcripts on one machine, all three cases above happened in chat and none reached a post.

So the automatic check watches the chat, and documents are covered by the passes that were going to read them anyway.

## `hooks/novel-words.py`

`Stop`. Reports words in the turn just finished that nobody has used earlier in the session, and blocks the turn until they are answered for.

A substituted name is by construction new to the session and the name it replaces is old, so the candidates need no model. A dictionary would not help: `rescue`, `crossings` and `residue` are all ordinary English, which is why they slip past.

A word is reported only if used at least twice in the turn, and plurals and verb forms of words already used are ignored — the first live firing was on `spawns`, with `spawn` already in the session. As configured it fires on **3.1%** of turns over 3,622 turns in three real sessions, median one word, against about a third of turns if every new word were reported. A third of turns is a message that gets skipped past, which is the failure this exists to avoid.

It is a partial net: of two known renamings in those transcripts it catches one. `hooks/README.md` says so rather than claiming coverage it does not have.

## What counts as already in use

Anything the user wrote, or the agent wrote in an earlier turn, counts on first sight. Tool output has to count for something too, or every term picked up from a file would look new — a column called `crossings` makes "the crossings" the data's own name, not an invention.

But a single appearance in tool output establishes nothing, and taken at face value tool output is mostly base64 and hex. So a word from it must appear at least twice: a real name recurs, a column heading in every row and a function at every call site, while base64 fragments are unique by construction. On a 62 MB transcript that is a 114,000-word vocabulary instead of 934,000, with `crossings`, `residue`, `partis`, `conscount`, `subagent` and `termcheck` all surviving.

## Cost, log and cleanup

The vocabulary is cached per session with the byte offset already read, so each turn parses only what is new: **0.10s** a turn and a 774 KB cache on that transcript, against 3.2s with no cache.

Every firing appends time, session, working directory, words and vocabulary size to `~/.claude/termcheck/firings.jsonl`, so leaving this running for a few weeks produces data rather than an impression. Silence is not logged. Caches for sessions untouched for a week are removed on each run.

## `hooks/terminology.sh`

`UserPromptSubmit`, restating the rule at the end of the context window each turn. It carries no example words: the text is an instruction to the agent reading it, and a list of words reads as a list of banned words — an earlier version listed `residue`, the correct domain term in one of the projects it fires in.

This is the less proven of the two. On the day it was written it did not prevent three renamings by the agent that wrote it. It is kept because it costs almost nothing and is the only part acting while text is being written rather than after.

## The rule in three skills

- **`bip-remove-metaspeak`** gets a "One name per thing" rule, and reports every name it put back alongside the passages it cut. This is the pass already run on issues and PR bodies before they ship.
- **`bip-ms-sweep`** Scan 6 is already called terminology consistency but covered only variant spellings and hyphenation. It now also covers one thing under two names, checking captions, axis labels, table headers and column names against the running text, since those drift independently of it.
- **`bip-issue-check`** gets it under internal consistency, where the concrete harm is a worker reading two names as two separate requirements.

Each already reads the whole document, so the rule adds no pass and no latency.

## Hooks are copied, not symlinked

Unlike agents and skills. A symlink into this repo dangles whenever another branch is checked out, and a hook whose command is missing exits 127, which Claude Code treats as non-blocking — the check would stop running with nothing said. `make test-hooks` runs 21 tests.

## What this does not do

`novel-words.py` reports candidates. It cannot tell whether a new word names something that already has a name; that judgement stays with the agent. It watches chat only, so a renaming that first appears in a file or a post is caught by the skills or not at all.

Things tried that did not work — a blocklist, a dictionary filter, example words in the reminder, instruction without detection, a subagent reading documents, and a hook gating GitHub posts — are recorded in #213 so they are not built again.


psathyrella:
@matsen ok I think this is working pretty well. Not sure if you'd rather merge it, or have other people test it on a separate branch. No real way to measure how well it's owrking other than that subjectively I think I'm getting way less novel/renamed jargon (I think mostly from restating the rule on each turn); and when the hook does occasionally fire, it's often useful and never causes a problem (the agent has no trouble identifying whether the hook is correctly pointing out renamed jargon).

PR #217, opened 2026-09-05
Port the remaining prose self-check items into scientific-tex-editor

psathyrella:
Folds the five items from `matsengrp/tex-template`'s prose self-check that this agent does not already cover: clefts, naming the actor, inert main verbs, the 40-word split threshold, and plain register.

Context is matsengrp/bcon-tex#3. The writing guidance is currently spread across three places that drift independently: the [writing_with_erick wiki page](https://github.com/matsengrp/wiki/wiki/writing_with_erick), a checked-in `misc/writing_with_erick.md` copy in every `-tex` repo, and `CLAUDE.md` in both `tex-template` and each paper repo. This agent is the one of them that actually gets iterated on, so it should be the place the style lives.

The other three template items are already covered by the flourish and short-sentence rules from #215.

One conflict resolved in favor of this file: the template said an em-dash pair is fine when it interrupts with a list or gloss, which contradicts the **Never use em-dashes** rule here.

Follow-up in the paper repos, not this PR: replace the restated style prose in `tex-template`'s and `bcon-tex`'s `CLAUDE.md` with a short pointer to this agent, and stop syncing the `misc/writing_with_erick.md` copies.

https://claude.ai/code/session_01TTpuB1uhVcnM5bSPjrkZJy


matsen:
This is great, thanks duncan!

PR #224, opened 2026-09-07
bip-pr-land: preserve .epic-worklog.md before deleting it at land time

matsen:
🤖

## Summary

`bip-pr-land`'s Step 6a now copies `.epic-worklog.md`/`.epic-status.json`
to `$CLONE_ROOT/.preserved/<issue>-<date>/` (with a provenance README and
a `gh pr comment` pointer) before anything can delete them, instead of
deleting unconditionally with no backup. `bip-conductor-poll`'s reclaim
rm gets the identical guard as a backstop, and `bip-conductor`'s
housekeeping prose is corrected to match.

The old Step 9.5 was the confirmed, unguarded deletion site for
matsengrp/phyz#2216: on 2026-09-06 it destroyed the deliberation record
for matsengrp/phyz#2314 (landed as matsengrp/phyz#2316, a `src/` fix to
default search behaviour in that repo — six source files, six re-pinned
regression tests, four phases of consumer triage, two independent agent
reviews). Gitignored, no backup, unrecoverable.

The premise that excused this ("a landed PR's work is in git, so the
worklog is redundant") is false: `PROSE-DISCIPLINE.md` has PR bodies
rewritten to current state *by design*, so deliberation deliberately does
not live there, and `issue-lead`'s periodic PR comments are evaluation-stop
summaries, not the worklog's continuous narrative. On a landed PR the
worklog is one of only two places the reasoning behind a change survives —
the PR/issue comment thread being the other, which lives on GitHub, not in
git objects a plain `git clone` would carry.

## Why Step 6a, not Step 9.5

Preservation has to happen before Step 8, not after: `bip worktree remove
--force` deletes the entire worktree directory, gitignored files
included, so a worktree-mode land would have lost the worklog at Step 8
regardless of anything done at the old Step 9.5 — one step earlier than
the bug this PR fixes, not actually fixed by the first pass. Step 6a
preserves from `$LAND_DIR` (captured before any `cd` or removal) in both
modes uniformly; Step 9.5 is now just the final `rm` plus landing a
`gh pr comment` pointer to the preserved copy. (Named "6a" because it
always ran between Step 6 and Step 7 — the old "7a" name misled about
where it sits relative to Step 8.)

Preserving in the same step that deletes also closes a *timing* gap a
reclaim-time preservation plan couldn't: the conductor's existing
gate-time workaround only protects a land that goes through a
request-gate, not every land.

## Blast radius

This skill is a symlink every session on the machine picks up the moment
it merges, including plain non-EPIC repos with no `.epic-config.json`.
The preserve block is gated on both `.epic-status.json` and
`.epic-config.json` being present, not just the former, so a non-EPIC
land sees nothing here rather than a scary `PRESERVATION FAILED`. A repo
that has `.epic-config.json` but still can't resolve `clone_root` from it
is a genuinely broken EPIC clone, and that case keeps the loud report.

## Review

Two rounds of `clean-code-reviewer`/`code-reuse-reviewer` on the diff,
plus the conductor and epic sessions on the joint landing gate, found and
fixed: `cp`'s exit status was discarded while the next line claimed
success unconditionally; `bip-pr-land` hand-rolled clone_root logic
`skills/lib/spawn-intent.sh`'s `resolve_clone_root()` already provides;
neither preserve block landed an in-repo pointer (`bip-conductor-poll`'s
own three-part rescue rule requires one — now posts via `gh pr
comment`/`gh issue comment`); and the blast-radius/naming points above.

Also checked: matsengrp/phyz#2216 flagged `walnut`/`hickory`/`aspen`/`beech`
missing `.epic-config.json`. All four already have the symlink (dated
2026-09-03) — fixed independently of this PR, nothing to do here.

## Recursion note

This bug was still live against this PR's own land until it merges.
Mitigated: copied this session's `.epic-worklog.md`/`.epic-status.json` to
`~/re/pz/.preserved/2216-stop-deleting-epic-state/` by hand before
requesting the landing gate.

## Test plan

No test harness exists for these skill markdown files — they are agent
instructions, not executable code. Verification here is a read-through for
internal consistency (Step 6a cross-references Step 8/9.5; the reclaim
backstop matches Step 6a's failure handling) plus two independent agent
review rounds and the joint conductor+epic gate review above.

Addresses matsengrp/phyz#2216. The tracking issue is deliberately kept on
`matsengrp/phyz` rather than moved here: that's where the damage was
observed and where EPIC matsengrp/phyz#369's evidence trail lives, while
this fix lands where the fleet tooling actually lives. No auto-close for
the same reason — this PR merges into `matsen/bipartite`, a separate repo
from the tracking issue, and this repo's history has no precedent for
cross-repo closing keywords; closing matsengrp/phyz#2216 is left to the
issue-lead/conductor once this merges.


matsen:
🤖 **Issue-lead follow-up on the merged change** — full assessment at matsengrp/phyz#2216 (reopened).

One defect in what landed here, plus one gap:

1. **`$PRIMARY` and `$PRESERVED_DEST` now cross Bash-tool invocation boundaries, where shell variables do not persist** (measured: set in one call, empty in the next; cwd does persist). This PR moved `cd "$PRIMARY"` out of the block that assigns it (`-    cd "$PRIMARY"`) into its own command, turning a working intra-command use into a cross-command one — so Step 6a will not `cd` in worktree mode. The rationale given for the split (Step 8 needs `$LAND_DIR` to resolve) does not hold; `$LAND_DIR` is absolute, from `pwd -P`. Worse because silent: Step 9.5's `gh pr comment ... \`$PRESERVED_DEST\`` posts an empty path, losing the discoverable pointer the three-part rescue rule requires. Round 2 cleared this as matching the pre-existing `$PRIMARY`/`$BRANCH` convention — but pre-merge, `$PRIMARY` was consumed inside the command that set it.

2. **Success criterion 3 (no-PR / stand-down path) has no executable guard.** The new `bip-conductor-poll` block sits under `#### Slot cleanup for merged PRs` and is gated on a merged PR; `bip-conductor-spawn` Step 2's fresh-assignment `rm -f .epic-status.json .epic-worklog.md` (SKILL.md:147, :164) is untouched. Mirror the same preserve-if-present block there.

Verified working, for the record: Step 6a's block dry-run against a synthetic clone preserves correctly, and the `.epic-config.json` gate correctly no-ops for a non-EPIC repo carrying a stray `.epic-status.json`.

PR #225, opened 2026-09-07
bip-pr-land/bip-conductor-*: preserve_epic_state() helper, fix 4 rounds of real bugs on top of #224

matsen:
🤖

## Summary

`bip-pr-land`, `bip-conductor-spawn`, and `bip-conductor-poll` now preserve `.epic-status.json`/`.epic-worklog.md` correctly in every case this issue's success criteria cover, via one shared `preserve_epic_state()` helper in `skills/lib/spawn-intent.sh` instead of four independently hand-rolled (and independently buggy) copies. Sits on top of matsen/bipartite#224, which fixed the original unguarded deletion at PR-land time but left a regression and a gap that five rounds of review (issue-lead, two automated reviewers, and the conductor and epic sessions reading these skills as their own daily executors) found and this PR fixes.

## What changed

- **`bip-pr-land` Step 6a** (worktree/land-time preservation) runs as one continuous Bash invocation — detect worktree mode, preserve, post a `gh pr comment` pointer, then `cd` to the primary clone if applicable — instead of splitting the `cd` and the pointer post into separate commands. Cross-invocation persistence is thread-type dependent (an agent/subagent thread resets both cwd *and* shell variables between separate Bash calls, measured directly; only a main interactive session's cwd survives), so anything that must carry state across what were separate commands has to live in one script now.
- **`bip-conductor-spawn` Step 2** (fresh-assignment and same-issue-restart cleanup, both clone- and worktree-mode) preserves before deleting, closing this issue's success criterion 3 ("a slot that reaches needs-human or stands down retains its worklog") — previously met only by prose, with the actual `rm` still unconditional at the exact site the issue named. Cleanup now uses `find <absolute-path> ... -delete` instead of `cd` + a separate relative-path `rm`: no `rm`/`rmdir` token in the command at all, so it can't trip Claude Code's destructive-removal guard regardless of `$` elsewhere in the invocation, and it doesn't depend on cwd surviving into a later command — so preserve and delete run together in one script.
- **`skills/lib/spawn-intent.sh`** gains `preserve_epic_state(source_dir, clone_root, reason)`: copies both files (if present) to `$clone_root/.preserved/<issue>-<date>/` with un-dotted, clone-namespaced filenames (`i<issue>-<clone>.worklog.md`/`.status.json`, matching this repo's existing convention — 23 of 25 prior entries use this form), writes a provenance README, and never silently overwrites an existing preserved copy for the same issue/clone/day (picks the next unused numeric suffix instead). Returns 0 (preserved, path on stdout), 1 (nothing to preserve, not an error), or 2 (status file present but uncopyable — caller must not proceed to delete the originals).
- **`bip-conductor-poll`'s reclaim backstop** now calls the same helper, as defense in depth for a land that somehow bypasses `/bip-pr-land`.

## Why this needed five rounds

`bip-pr-land`, `bip-conductor-spawn`, and `bip-conductor-poll` are symlinked live into `~/.claude/skills/*` from this repo's working tree — verified: `readlink ~/.claude/skills/bip-pr-land` → this repo's checkout. Every Claude session on the machine executes whatever is currently checked out here, branch included, the instant it runs `/bip-pr-land` or `/bip-conductor-spawn` — not `main`, and not only after merge. That's also why the four independently-drifting preserve implementations this PR consolidates were worse than an ordinary DRY violation: each was independently live and independently wrong, in a different way, at a different time, to the whole fleet. It is also why holding this PR back doesn't reduce exposure — the code has been executing all along regardless of review status; landing reviewed, fixed code replaces that exposure with something someone actually checked, rather than creating a new one.

Given that, this went through five review rounds, each surfacing a genuine defect in a different category: variable scoping, directory-resolution ordering, error masking, code duplication, guard-word phrasing, and a same-day preserve collision. Every defect found was real, reproduced directly against a synthetic clone or the checked-out helper itself (not dry-run), and is fixed in the current diff. The full commit history carries the round-by-round chronology (five commits, one per round); it isn't repeated here per this repo's own `PROSE-DISCIPLINE.md`. Whether that pattern reads as convergence or as a change surface still being characterized is left to the reader — both the conductor and epic sessions reviewed every round and can speak to it directly.

## What's verified, and what isn't

Verified directly against synthetic clones and the live helper: full preservation success; the no-op path (nothing to preserve); the partial-success path (status present, worklog legitimately absent — preserves the status file and notes the gap, rather than failing outright); a genuine hard failure (status file present but uncopyable — returns failure and cleans up the empty destination); the same-day/same-issue collision case (both worklogs survive under distinct names); Step 6a's exit status in both clone and worktree mode; and every fenced bash block in the three touched skill files audited for any remaining single-command `$`+`rm`/`rmdir` combination.

Since verified against a real land: `/bip-pr-land`'s Step 6a fired end to end during matsen/bipartite's own review of this PR, landing an unrelated issue's PR (i2294, matsengrp/phyz#2323) — `.preserved/2294-<date>/` was created with the un-dotted, clone-namespaced filenames, a correct README, and a posted `gh pr comment` pointer, exactly as designed. That observation is evidence about *this* PR, not about #224 (which does not call the new helper).

One real fragility surfaced by the same night's activity, not fixed here: an earlier land (i2321) fell back to hand-copying instead of calling `preserve_epic_state()`, because `source "$(dirname "<this-skill's-base-directory>")/lib/spawn-intent.sh"` failed to resolve — most likely the `<this-skill's-base-directory>` placeholder wasn't substituted with a real path before the command ran. That placeholder-substitution pattern is a pre-existing, repo-wide convention (used by `resolve_clone_root` and `find_spawn_intent` well before this PR), not something this PR introduced, but this PR does make all four preserve sites newly dependent on successfully sourcing one shared file — a failure mode the four independent hand-rolled copies didn't have. Worth hardening in a follow-up (e.g. a well-known absolute fallback path, or a louder failure than a silent fallback to hand-copying) but out of scope here since it isn't specific to this issue's fix.

Still not verified: a real `/bip-conductor-spawn` invocation calling `preserve_epic_state()` during an actual respawn (the fresh-assignment and same-issue-restart paths).

## Deferred

> **Deferred**: this PR's own branch lives in `~/re/bipartite`, the same clone `~/.claude/skills/*` symlinks into, so the fleet has been executing this branch rather than reviewed `main` for the duration of its review — a structural risk independent of this PR's content, and true of every bipartite PR ever developed from this box, not specific to this one. The safer alternative isn't "park on `main` while this is pending" (verified: `main` still carries #224's two defects this PR fixes, so the branch checked out now is strictly the safer of the two states). The actual fix is process: fleet-tooling development should happen in a separate clone or worktree per branch, with `~/re/bipartite` itself parked on `main`. That's infrastructure work distinct from this issue's scope and needs its own issue.

## Test plan

See "What's verified, and what isn't" above.

Addresses matsengrp/phyz#2216 (reopened after #224). No auto-close: this PR lands in `matsen/bipartite`, a separate repo from the tracking issue — deliberately: #2216 is filed where the damage was observed and where EPIC matsengrp/phyz#369's evidence trail lives, while the fix lands where the fleet tooling actually lives.

## Landing note

This PR touches shared fleet tooling (`bip-pr-land`, `bip-conductor-spawn`, `bip-conductor-poll`, `spawn-intent.sh`), which EPIC matsengrp/phyz#369's landing-authority ruling carves out of delegated two-session (conductor+epic) approval — it needs the user's sign-off directly. Both the conductor and epic sessions reviewed this PR's substance in full, independently, across all five rounds, and neither is withholding on content — only on authority.


matsen:
🤖 **Issue Lead** (iteration 2) — evaluating matsengrp/phyz#2216

**Category**: quality-gate (not terminal — one blocker remains before this is ready for the user's landing decision)
**Scope check**: on-track. No drift. The one DEFERRED item (this branch living in the clone the fleet symlinks from) is genuine follow-up work, not a punt — it needs a change to how fleet-tooling development is structured, not another edit to these four files.

## The four rounds are real, and I checked them rather than believing them

I sourced the checked-out `skills/lib/spawn-intent.sh` and drove `preserve_epic_state()` directly. Every claim in the body holds:

- full success — both files, un-dotted clone-namespaced names, `rc=0`
- partial (status present, worklog absent) — `rc=0`, status file preserved, README notes the absence. This is round 3's `&&`-chain regression, genuinely fixed.
- no-op (no status file) — `rc=1`, silent
- hard failure (status present, uncopyable) — `rc=2`, destination `rmdir`'d, nothing left behind that reads as a rescue

Step 6a's `if [ -n "$PRIMARY" ]; then cd "$PRIMARY"; fi` returns 0 in clone mode. Round 3's inverted-exit-status bug is really gone.

Four rounds of real bugs is the gate working. I agree with the conductor on that.

## Blocker: a fifth defect, and it is this issue's own failure class inside the remedy

`preserve_epic_state()` keys the destination on issue and date, with fixed filenames:

```sh
dest="$clone_root/.preserved/$issue_n-$(date -I)"
cp "$source_dir/.epic-worklog.md" "$dest/i$issue_n-$clone_name.worklog.md"
```

A second preserve for the same issue, from the same clone, on the same day silently overwrites the first. Measured:

```
A. preserve worklog "WORKLOG-ATTEMPT-ONE"           -> rc=0
B. preserve worklog "WORKLOG-ATTEMPT-TWO..."        -> rc=0, same dest
   grep -rl WORKLOG-ATTEMPT-ONE .preserved/         -> no match
```

No warning, `rc=0` both times, first worklog unrecoverable. That is silent, unrecoverable worklog loss — the thing #2216 exists to prevent — now living in the function written to prevent it.

Two live paths reach it:

- **`bip-pr-land` Step 6a firing twice for a reopened issue that lands two PRs in one day.** That is this issue's own shape: #224 landed, #2216 reopened, #225 follows the same day. Step 9.5 deletes the worklog after land 1, so land 2's worklog is a *new* file, not a superset — the first is genuinely lost, not merely rewritten.
- **`bip-conductor-spawn`'s worktree-mode block**, whose stated premise is "a fresh restart on the same issue."

Fix belongs in the shared helper, where all four call sites inherit it: refuse to clobber an existing `$dest` (suffix `-2`, `-3`) or add a time component. Roughly five lines.

## The cwd contradiction is resolved — it shouldn't ship as an open unknown

The body carries this as "Separate observation, not acted on … this one measurement may not generalize." It generalizes, and the rule is simple: **persistence is thread-type dependent.** Measured this session in an agent thread — `cd` to a scratch dir, next call reports the original cwd; a variable set in one call reads `EMPTY` in the next. Agent threads reset both. Main-session threads persist cwd. That reconciles the two contradicting measurements: both were right about their own thread type.

This matters because two prose claims in the files this PR edits are now known false, one of them a sentence this PR introduced:

- `bip-pr-land/SKILL.md:120` — "Shell variables do not survive between separate Bash invocations (only the working directory does)"
- `bip-conductor-spawn/SKILL.md:210` — "The Bash tool's working directory persists between calls"

Worth folding in while correcting them: `bip-conductor-spawn`'s Step 2 does `cd "$SLOT"` in one command and `rm -f .epic-status.json` in the next, so in an agent thread the delete runs in the wrong directory. The repo's own documented guard-safe idiom fixes it and is cwd-independent — `find "$SLOT" -maxdepth 1 -name '.epic-status.json' -delete`, which cannot trip the destructive-removal guard because the guard short-circuits without an `rm`/`rmdir` token. That removes the last cross-invocation dependency from the exact deletion sites this issue is about.

## Rewrite the body as current state before the user reads it

`PROSE-DISCIPLINE.md` names the "Round 0 / Round 1 / Round 2 / Round 3" structure specifically: revision narrative belongs in commit messages, which here already carry every round in full. The user is the landing decider and the least-context reader in the chain; they should get what the four files now do, what is verified, and what isn't — not a reconstruction exercise. Keep and promote the two facts that actually drive the decision: symlink-live execution, and the honest limit that nothing here has been exercised by a real land or a real respawn.

## On the landing gate itself

Unchanged, and I'm not adding a second one. The shared-tooling carve-out is the user's call. The items above are work to finish *before* that call, not another approval.

One thing the escalation should say out loud: **this branch is checked out in `~/re/bipartite` right now, which is the clone `~/.claude/skills/*` symlinks into — so the fleet is executing this branch, not `main`, while the user sleeps.** "Park on main until approved" is the wrong instinct: `main` still has #224's empty-pointer and clone-mode-exit-1 defects, so it is strictly worse. Leaving the branch checked out is the safer option, and the user should be told that rather than left to infer it.

**Action**: continue at quality-gate — fix the same-day overwrite, correct the two cwd claims, rewrite the body — then re-park at `needs-human` and re-notify. Do not land.

matsen:
🤖 **The "not verified" item in this PR body now has a partial answer, and it is not a clean pass.** Recording it here because it bears on the decision, and because this PR is the thing that introduces the dependency.

**`preserve_epic_state()` still has not executed in a real land.** matsengrp/phyz#2321 landed and Step 6a's pointer-comment half fired, but the helper half did not — the worker fell back to hand-copying. **The outcome was correct** (right directory, right `i<issue>-<clone>` naming, byte-identical, diff-verified). Nothing surfaced the fallback except reading its own footnote.

**The recorded cause was wrong and the real one is a property of this PR's design.** The fallback was attributed to `skills/lib/spawn-intent.sh` being *"missing on this machine."* It was present throughout — both resolutions exist, and sourcing it makes the function callable:

```
$ ls -la ~/.claude/skills/lib
lrwxrwxrwx ... -> /home/matsen/re/bipartite/skills/lib
$ [ -f ~/.claude/skills/lib/spawn-intent.sh ]        # EXISTS  (mtime 20:18; PR merged 20:57)
$ bash -c 'source ~/.claude/skills/lib/spawn-intent.sh && type -t preserve_epic_state'
function
```

It was a **resolution** failure. The source line is:

```
source "$(dirname "<this-skill's-base-directory>")/lib/spawn-intent.sh"
```

Left unsubstituted, `dirname` of the literal placeholder is `.`, so it resolves against the worker's **clone**, not the skill directory:

```
$ cd ~/re/pz/alder && dirname "<this-skill's-base-directory>"   =>  .
$ [ -f ./lib/spawn-intent.sh ]                                   =>  MISSING
```

**An unsubstituted placeholder is indistinguishable from a missing file at the point of failure.** `[ -f ]` is false either way, which is exactly why the wrong cause got written down.

**Why this is a finding about the design and not just an incident.** This PR's central improvement — consolidating four independently-drifting preserve implementations into one shared helper — is the right call, and I'd say so plainly. But the four hand-rolled copies had one property the helper does not: **they could not fail to load.** They were inline. The helper must be `source`d through a line carrying a placeholder that an agent has to substitute correctly, in a repo where cwd is a worker's clone rather than the skill directory. All four call sites now inherit that. It trades a drift failure mode for a load failure mode, and the load failure is silent, mis-diagnoses as a missing file, and is only visible if someone reads the fallback's footnote.

**This is not an argument against landing.** The fallback worked, the outcome was correct, and the drift the helper eliminates was demonstrably worse. It is an argument for hardening the resolution before the next four call sites rely on it — an absolute path, or a resolution that fails loudly and distinguishably from "file absent" rather than falling through to `.`.

matsengrp/phyz#2321's comment thread carried the same wrong cause and has been corrected there. The next real land (matsengrp/phyz#2320) is the remaining test of whether the helper ever fires.

matsen:
🤖 **Four real lands have now exercised this. The helper works when it runs — it ran in 2 of 4 — and both failures were in the `source` line, not in `preserve_epic_state()` itself.**

Verified directly against `$CLONE_ROOT/.preserved/`:

| land | helper ran? | destination |
|---|---|---|
| p2321 (alder / i2312) | **no** | `2312-warm-start-gtr/` — hand-rolled fallback, made its own directory |
| p2323 (elm / i2294) | **yes** | `2294-2026-09-06/` |
| p2324 (ash / i2318) | **yes** | `2318-2026-09-06/` |
| p2320 (teak / i2313) | **no** | **loose in `.preserved/` root** — no directory at all |

```
$ ls ~/re/pz/.preserved/ | grep -v /
i2313-teak.status.json
i2313-teak.worklog.md
README.md          <- the directory's own README, now sitting alongside them
```

**The success path is sound.** Both runs produced the intended `<issue>-<date>/` directory with the un-dotted `i<issue>-<clone>` filenames. That closes this PR's one honest unverified claim — the helper had never run against a real land when the body was written, and now it has, twice, correctly.

**The failure mode is the one worth weighing, and four data points sharpen it.** Neither failure was a defect in the helper; both were callers falling back because the `source "$(dirname "<this-skill's-base-directory>")/lib/spawn-intent.sh"` line did not resolve. Left unsubstituted, `dirname` of that literal placeholder is `.`, which resolves against the worker's clone rather than the skill directory — and `[ -f ]` is then false in a way indistinguishable from the file being absent. That is why the first failure was written up as "missing on this machine" when the file was present the whole time (corrected at matsengrp/phyz#2321).

**The specific thing I'd put in front of a reviewer: extracting the shared helper removed drift from the success path and left it in the fallback path — and the fallback path ran half the time.** Three lands, three different destinations:

- `2294-2026-09-06/` (helper)
- `2312-warm-start-gtr/` (alder's hand-rolled fallback)
- `.preserved/` root, no directory (teak's hand-rolled fallback)

The *filenames* converged — both fallbacks produced correct `i<issue>-<clone>.{status.json,worklog.md}` — so the naming convention propagated through the docs. The *directory* convention did not. Four independently-drifting preserve implementations were the defect this PR set out to fix; four independently-drifting **fallbacks** are what it leaves behind, and they are reached whenever the `source` line fails.

**None of this argues against landing.** The consolidation is right, the helper is correct, and the drift it eliminates was demonstrably worse. It argues for hardening the resolution before more call sites depend on it — an absolute path, or a resolution that fails loudly and *distinguishably* from "file absent" rather than falling through to `.` and leaving each caller to improvise.

**One consequence for the loose files, recorded so nobody tidies them by reflex:** p2320's committed pointer comment names those root paths, so moving them would break the only durable record. They should be left where they are until the pointer is updated in the same change.

PR #247, opened 2026-09-17
Add /bip-pr-file: write the PR body before filing it

psathyrella:
Not sure I sign onto all of CC's prose below, but I have been finding that PR bodies are often pretty useless for figuring out what the PR actually does, so I'm trying to come up with guidelines to help CC write useful bodies. 

> Claude Code:

Adds `/bip-pr-file`, a skill for writing a PR body and filing the PR.

The skills cover every stage of a PR's life except the first. `/bip-pr-check`
runs a readiness check, `/bip-pr-review` runs the pre-merge checklist, and
`/bip-pr-land` merges — but creation, which is where the body actually gets
written, has no skill. Four other skills call `gh pr create` in passing, each
carrying its own one-line note about the body.

That gap produces bodies a reviewer cannot read. This came up when a user
reviewing one said "i'm trying to figure out why i still have no idea what it
does... so many PR bodies are just fragments of nonsense, instead of actually
describing what they do and why". The body in question spent 150 words on a
defect and never stated the change, which was one sentence. Checking the last
eight merged PRs in that repo, five have section headings that name only a
defect: "The check never fired", "Wrong numbers", "Silent success", "Two bugs".

## The skill leads with the reader, not the diff

A reviewer is by definition not involved in the details of the work. So the body
opens with a short paragraph of prose saying what the PR does and why, with
whatever background places it — no heading above it, no table, no bullets. Then
one section per change: what the problem was, what changed, why.

**Each section opens with one sentence naming the change**, subject-verb-object.
The check is mechanical: read the first sentence of each section in sequence and
see whether that describes the diff. A heading naming a defect rather than a
change is the tell, and the skill gives three before/after pairs.

A PR covering several unrelated changes still gets one opening paragraph.
Combining is explicitly not the problem — a body that fails to say what it does
is.

## One rule that is new

An approach that did not work is worth a sentence when it is an obvious
thing to try that cost real time to rule out and someone will try it again.
`/bip-remove-metaspeak` would cut that as process narration; it is worth an
exception, written as "what not to do and why" rather than as the story of
discovering it.

## Not in this PR: the hook

A skill only fires when invoked, and `hooks/README.md` on #214 makes the case
already — "stating the rule in `CLAUDE.md` does not hold it: the rule is read
once, at the start of a session, and a session runs for hundreds of thousands of
tokens after that". A `PreToolUse` hook on `gh pr create` would catch a PR filed
without the skill.

It is not here because `hooks/` does not exist on main — it arrives with #214,
which is open and nine commits ahead. Stacking on someone else's open PR is a
decision for its author, so this PR is the skill alone, which has no dependency.

`README.md`'s key-skills line and a pointer from `/bip-pr-check` are the only
other changes.

https://claude.ai/code/session_012sCHtgKCseW9fXzwP2vkyG


psathyrella:
> Claude Code:

Folded into #248, which is stacked on #214 as asked. This PR existed only because I opened it before checking where `hooks/` lived, which forced the split into two.

PR #248, opened 2026-09-17
Add /bip-pr-file and the hook that makes it fire at `gh pr create`

psathyrella:
Not sure I sign onto all of CC's prose below, but I have been finding that PR bodies are often pretty useless for figuring out what the PR actually does, so I'm trying to come up with guidelines to help CC write useful bodies.

edit: as with my other PR, this is experimental, I'll report back here when I figure out how well it's working.

> Claude Code:

Adds `/bip-pr-file`, a skill for writing a PR body, and `pr-body.py`, a hook
that makes it fire at `gh pr create`.

The skills cover every stage of a PR's life except the first. `/bip-pr-check`
runs a readiness check, `/bip-pr-review` runs the pre-merge checklist, and
`/bip-pr-land` merges — but creation, which is where the body actually gets
written, has no skill. Four other skills call `gh pr create` in passing, each
carrying its own one-line note about the body.

That gap produces bodies a reviewer cannot read. A user reviewing one put it as
"i'm trying to figure out why i still have no idea what it does... so many PR
bodies are just fragments of nonsense, instead of actually describing what they
do and why". The body in question spent 150 words on a defect and never stated
the change, which was one sentence. Across the last eight merged PRs in that
repo, five have section headings naming only a defect: "The check never fired",
"Wrong numbers", "Silent success", "Two bugs".

**Stacked on #214**, which is where `hooks/` and its README arrive. Base this on
that branch, not `main`.

## The skill leads with the reader, not the diff

A reviewer is by definition not involved in the details of the work. So the body
opens with a short paragraph of prose saying what the PR does and why, with
whatever background places it — no heading above it, no table, no bullets. Then
one section per change: what the problem was, what changed, why.

**Each section opens with one sentence naming the change**, subject-verb-object.
The check is mechanical: read the first sentence of each section in sequence and
see whether that describes the diff. A heading naming a defect rather than a
change is the tell, and the skill gives three before/after pairs.

A PR covering several unrelated changes still gets one opening paragraph.
Combining is explicitly not the problem — a body that fails to say what it does
is.

An approach that did not work earns a sentence when it is an obvious thing to
try that cost real time to rule out and someone will try it again.
`/bip-remove-metaspeak` would cut that as process narration; it is worth an
exception, written as "what not to do and why" rather than as the story of
discovering it.

## The hook makes the skill fire at the moment it is needed

A skill only runs when something invokes it, and the bodies that prompted this
were all filed by sessions that never opened one. `hooks/README.md` already
makes the argument for the terminology hooks: a rule stated in `CLAUDE.md` is
read once, at the start of a session, and a session runs for hundreds of
thousands of tokens after that. `gh pr create` is a precise moment; `CLAUDE.md`
is not.

`pr-body.py` is a `PreToolUse` hook on Bash. It blocks a `gh pr create`, or a
`gh pr edit` that replaces the body, when the transcript holds no `Skill`
tool-use record for `bip-pr-file`. Exit 2 means the command does not run and the
stderr text reaches Claude, which loads the skill and retries; the second
attempt finds the record and passes.

### It checks that the rules were loaded, not that the body is good

Those are different jobs, and the first is the one a hook does reliably.
Catching a PR filed with no body rules in front of the author is the whole of
the enforcement — a skill nobody opens at the moment it is needed changes
nothing.

What it cannot do is grade prose. The defect the skill exists for is sections
that read well and never state the change, and no regex sees that. The docstring
and the README say so, so that nobody later reads a passing hook as a passing
body.

### Three ways it stays quiet

- **The skill's name in prose is not evidence.** Only the `Skill` tool-use
  record counts. This hook's own message names the skill, so a looser match
  would pass on the retry with the skill never read. There is a test for that
  case specifically.
- **Nothing is blocked when the skill is not installed.** Blocking with no
  remedy to point at is worse than not firing.
- **An unreadable transcript does not block.** That is our failure, not the
  author's.

`gh pr view`, `list`, `merge`, `checks`, `edit --add-label` and `create --help`
are all untouched.

## Tests

11 for the hook, covering both directions plus each quiet case. `make
test-hooks` now runs both hook suites. `SKILL_DIR` is overridable through
`BIP_PR_FILE_SKILL_DIR` so the tests do not depend on what is installed on the
machine.

Like the other hooks it is copied rather than symlinked by `make install` and
has to be named in `~/.claude/settings.json` by hand — the README section gives
the entry.

`README.md`'s key-skills line and a pointer from `/bip-pr-check` are the only
other changes.

https://claude.ai/code/session_012sCHtgKCseW9fXzwP2vkyG


matsen:
Landed the spirit of this in #299, though not the implementation.

The two checks that would actually change a body — read the first sentence of each section in sequence and confirm it describes the diff, and the "heading names the change, not the defect" table — went into `/bip-pr-check` Step 6, which is the body rubric we already run (and which the conductor gate loop already invokes). Both conductors reviewed and independently landed on folding in rather than adding a fourth PR skill, so that's the form it took.

What didn't come along: the standalone `/bip-pr-file` skill, the `pr-body.py` hook (Python + its tests in an otherwise pure-Go repo, and stacked on #214), and the restatement of `PROSE-DISCIPLINE.md`/`EVIDENCE-DISCIPLINE.md` that `CLAUDE.md` already requires for every PR body. `conductor-spawn`'s PR-creation step now points at the rubric so worker bodies get it at creation.

Thanks — the core observation (bodies that read well and never say what changed) was the right one to catch.

PR #253, opened 2026-09-21
Port the file-overlap check to `bip epic collide`, and close its fail-open

### LINKED ISSUE #252: Port the file-overlap half of fleet-collisions.sh to `bip epic collide`
# Port the file-overlap half of `fleet-collisions.sh` to `bip epic collide`

🤖 After this issue, `bip epic collide` performs the live-clone file-overlap check in Go with a scope it derives rather than defaults, `skills/lib/fleet-collisions.sh` keeps only its pane/process sections, and `skills/bip-epic/SKILL.md`'s symbol blast-radius block is replaced by an invocation.

## Motivation

**The decisive reason is a live fail-open that has already authorised a spawn against the wrong fleet, and it is unreachable in Go.** `skills/lib/fleet-collisions.sh:59` is `ROOT="${1:-$HOME/re/pz}"`. `skills/bip-conductor/SKILL.md:529` records the status: *"`fleet-collisions.sh`'s own `ROOT="${1:-$HOME/re/pz}"` still has the original defect and is mitigated only by the explicit-argument instruction above; treat that as unfinished, not as fixed."*

Three properties compound it. The scope argument is optional; `${x:-default}` substitutes on **empty** as well as unset, so an unparseable `.epic-config.json` makes an explicitly-passed root evaporate; and the fallback is a *different repository's* clone pool. Measured 2026-09-17 on `matsengrp/superfamily-pcp` (`skills/bip-conductor/SKILL.md:720-734`): an argument-less run reported clones `birch`/`cedar` and `COLLISION src/ml/stochastic_search.zig`, while an explicit-root re-run in the same minute reported `cobalt`/`copper`/`silver` and `COLLISION CLAUDE.md`. Every clone name, branch and file differed, nothing in the first output looks wrong, and it missed a real three-way `CLAUDE.md` collision that went on to produce two conflicting PRs.

`skills/lib/clone-currency.sh` already fixed exactly this — it derives `clone_root` from `.epic-config.json` and prints the scope it resolved. This issue brings the same treatment to the overlap check, in a language where the defect cannot be spelled.

**Secondary: the prescribed guard against the other fail-open does not work in this fleet's shell.** The script's header remedy for reading a status after a pipe is `"${PIPESTATUS[0]}"`, which evaluates to the empty string in zsh — `pax`'s login shell (`skills/bip-conductor/SKILL.md:734`). And the `cmd | wc -l` class catalogued in `skills/lib/hazards/counting-a-failed-command.md` had an instance in this very script until `d2e0e10`.

**Tertiary: the procedure is duplicated as prose.** This is layer 3 of `docs/guides/corpus-capture.md`, the only layer that reduces load-time context. `skills/bip-epic/SKILL.md` Step 4b is 15,117 bytes (measured at `main` `9ee88de`, `sed -n '328,388p' skills/bip-epic/SKILL.md | wc -c`) with no fenced command block, though three runnable commands appear inline — the shape that has no `--help` to drift from. Only ~2,494 bytes of it is procedure; the rest is judgement and stays. See Out of scope.

## Scope — read this before implementing

`fleet-collisions.sh` emits **five** report sections. Sections 1, 2 and 2b are file overlap and git state. Sections 3, 4 and 5 are stale `.epic-status.json`, a live pane with no status file, and dead agent sessions — pane and process state needing `tmux list-panes`, a process-subtree walk (`pane_has_agent`), and `jq`.

**Only sections 1/2/2b are in scope.** The script is **not deleted**: it keeps 3/4/5. Those belong with `#238`'s `bip epic status` surface, not with a git-overlap command.

**Signature:** `bip epic collide [--root <clone-root>]` — one clone pool, every clone in it, the same universe as the script. It is not a two-ref tool. Grouped under `bip epic` because the pool is defined by `.epic-config.json` and the existing fleet commands live there (`cmd/bip/epic.go`); a top-level `bip collide` would sit among the reference-library commands.

## Behaviour to carry over

Each item cites its site. The ones carrying a measured incident say so.

- **Scope is derived, never defaulted** — the fix, not a port. Resolve the pool root from `.epic-config.json` (the key `resolve_clone_root` reads, `skills/lib/spawn-intent.sh:20`), print it as the first output line, and exit uncheckable if it cannot be resolved. An empty root is not a root.
- **Three-valued verdict: clear / found / could not check.** `fleet-collisions.sh:152`: *"Fail toward checking, never toward skipping."* Could-not-check dominates found, because a caller acting on *found* believes the pool was fully examined (`:117-130`).
- **Merge-base, not two-dot.** Measured 2026-09-14: a slot one commit behind reported 15 touched files having touched 6, and the 9 phantoms produced a false collision. Where `merge-base` fails the script falls back to `origin/main` (`:174`), silently reinstating the two-dot behaviour — make that **uncheckable** instead.
- **`status --porcelain -z` is the untracked-file pass only.** The merge-base `diff --name-only` already covers uncommitted *tracked* changes because it diffs the working tree against a commit (`:47-51`). NUL-delimited because git quotes paths containing spaces; `R`/`C` entries emit the old path as a second NUL field with no status prefix, and either path can collide.
- **Foreign clones are excluded, not reported** (`:131-156,237-244`). A pool root can hold clones of other repositories — the phyz pool holds a pinned checkout of `matsen/iqtree2`. It sat at a permanent detached tag commit, so it reported UNCHECKABLE on every run and made **every run exit non-zero**, *"destroying the signal the stickiness fix had been added to protect, one commit after adding it."* `is_pool_clone` returns **true** when the modal origin is unknown — fail toward checking.
- **`git rev-parse --absolute-git-dir`, never `[ -d "$d/.git" ]`** (`:246-255`). In a linked worktree `.git` is a *file*, so the directory test is false and both weird-state paths silently skip it — *"reinstating, in a supported mode, the exact gap they exist to close."* Worktree mode is opt-in in this repo (`flow.ResolveRepoPath`, `docs/guides/layout.md`), so the port will meet it.
- **Mid-rebase recovery from both backends** — `rebase-merge/` and `rebase-apply/head-name`; detached HEAD with no rebase is uncheckable (`:257-260`). *"The clone in the weird state is the one you most want checked."*
- **Empty-pool-root guard** (`:61-83`). Measured 2026-09-14: globbing an empty directory errors in zsh and does one silent iteration on a nonexistent path in bash. A real pool root holds `.spawn-prompts/`, `.preserved/`, `.mirror/`, so the dotfile filter is load-bearing — *"the guard and the glob must share a universe."* Exit clear with "nothing to check", not clear.
- **A failed `git fetch origin main` warns and continues** (`:168-169`), because `origin/main` being stale produces false collisions rather than missed ones. Stated so it is a decision rather than a rediscovery.
- **Printed denominators** — `clear (mine=36 landed=44)` (`:229-232`). Without them a trivially-clear result, a genuinely-clear result and a broken check all print the same word.
- **Capture-then-count, never `cmd | wc -l`**, and the failure sentinel must not compare equal to `0` downstream (`skills/lib/hazards/counting-a-failed-command.md`).

## Implementation

**Phase 1 — the overlap check.** `cmd/bip/epic_collide.go` plus logic in `internal/gitx/`, beside the existing `worktree.go` whose doc comment is *"Package gitx provides small, tested wrappers over the git porcelain"*. Not `internal/flow/` — that is GitHub-activity state (`internal/flow/types.go:1`) and `#39` proposes renaming it. Not `internal/git/` — its diff/log surface is typed in `reference.Reference` (`internal/git/types.go:6`); its generic helpers `FindRepoRoot`, `IsGitRepo` and `ValidateCommit` can be called without living there. Avoid the name `conflict`: `internal/conflict/` is refs.jsonl conflict resolution. Cobra idiom per `cmd/bip/diff.go`: `func init() { rootCmd.AddCommand(...) }` above a package-level `var` with `RunE`.

**Exit codes need a decision in this PR, and the existing table forces it.** `cmd/bip/exitcodes.go` already binds `1 = ExitError` and `2 = ExitConfigError`, and `cmd/bip/main.go:23-28` maps *every* `RunE` error to `ExitError`, so a distinct "could not check" cannot arrive through a returned error. Add `ExitCollisionFound = 10` and `ExitCollisionUncheckable = 11` — new values rather than overloading `1`/`2`, whose current meanings contradict the contract — have `RunE` return `nil` on every verdict and call `os.Exit` directly, and update the `# exit 0 = clear, 1 = found, 2 = could not check` comment at `skills/bip-conductor/SKILL.md:717` to the real values in the same PR.

Same PR: rewrite `skills/bip-conductor/SKILL.md:708-734` to run `bip epic collide` and then the script's remaining sections, remove sections 1/2/2b from the script, and paste `bip epic collide --help` into the skill's invocation block. `CONSTITUTION.md:27-34`: *"Skill docs are canonical — when CLI behavior changes, skill docs MUST update in the same PR"* and *"Skill behavior MUST match `--help` output exactly."* These skills are symlinked live, so a command-only landing leaves running sessions on a superseded procedure.

**Phase 2 — `--symbol`.** `git grep -ln <SYMBOL>` blast radius, reporting **code hits and prose hits as separate counts**, plus a third line naming what a grep cannot see: files importing the defining module without naming the symbol. Same PR replaces `skills/bip-epic/SKILL.md:363` and all four bullets of the list opened at `:365` (i.e. `:365-369`) with the invocation plus the part the command cannot supply — the commit and date the count was taken at. `:369` must be carried, not dropped: measured at three commits, 14 indirect consumers held steady while the direct count moved 26→39, and *"a worker who greps gets the direct set and believes it has the set."* Shipping `--symbol` without that hedge reproduces the defect `:369` names.

**Out of scope:** `skills/bip-epic/SKILL.md:343-362` and `:370-387` — scope, hand-off, citation-collision and reporting-vocabulary. Judgement, not procedure. Also out of scope: merge-conflict detection. The script does none, and `git merge-tree --write-tree` per pair is prescribed at `:379` as the conductor's side of the hand-off; it can be a follow-up, but overlap is the deliverable here. `:379` also records why it would not suffice alone — two PRs re-pinned different rows of one file, git merged them cleanly, and whether the citations resolved was decided by a separate check: *"a clean auto-merge is not a correct merge."*

## Test plan

Against real git fixtures. Reuse `initRepo(t *testing.T)` from `internal/gitx/worktree_test.go`, which pins `user.name`, `user.email` and `init.defaultBranch` locally so fixtures do not inherit developer git config.

1. **Merge-base, not two-dot.** `main` = M1→M2→M3; `A` forks at M1 touching only `a.txt`; `main` gains `c.txt` at M2, `d.txt` at M3; `B` forks at M3 touching only `b.txt`. Expect `A` = `{a.txt}`, no overlap, clear. A two-dot implementation returns `{a.txt, c.txt, d.txt}` and fires a false collision.
2. **Rename old path and space in path.** `A` does `git mv x/common.py x/shared.py`; `B` edits `x/common.py`; both edit `docs/a b.md`. Expect overlap `{x/common.py, docs/a b.md}`, found.
3. **Untracked file collides.** `A` and `B` each create an uncommitted, untracked `scratch/notes.md`. Expect overlap — this is the case the `--porcelain -z` pass exists for and `diff --name-only` cannot see.
4. **Degenerate refs are uncheckable, not clear.** A clone with no commits relative to its merge base, and a pool of one clone. Neither is a clear verdict nor a 100%-overlap report.
5. **Unreadable clone is uncheckable, and the count is not `0`.** Assert the reason is named, that stdout carries a non-numeric sentinel, and that it does *not* match `=0`. Assert uncheckable wins when another pair reports found.
6. **Denominators print on a clear result.** Base `B`; `origin/A` adds three files off `B`; `origin/main` adds none. Expect `mine=3 landed=0`, clear. Then `origin/main` gains two past `B`: expect `mine=3 landed=2`. Catches an implementation that prints `clear` with no numbers.
7. **A foreign clone is excluded, not reported.** Pool with two clones of origin `X` on live branches and one clone of origin `Y` at a detached tag. Expect the `Y` clone absent from all output and a clear verdict. Catches the regression where one permanently-detached foreign clone made every run exit non-zero.
8. **A linked worktree mid-rebase is checked, not skipped.** `git worktree add` so `.git` is a file, then a conflicted rebase using the `--apply` backend so state lives in `rebase-apply/head-name`. Expect the branch name recovered and the comparison run.
9. **Symbol split.** A symbol in both `.go` and `.md` reports two counts, not their sum.
10. **An empty pool root, and one holding only `.spawn-prompts/`.** Both report "nothing to check" and neither reports clear.
11. **Scope is printed and never defaulted.** With no `--root` and an unparseable `.epic-config.json`, expect uncheckable — not a scan of any other directory. Assert the resolved root appears in stdout on every successful run.

## Success criteria

1. Every item in "Behaviour to carry over" has a test that fails against an implementation lacking it.
2. `skills/lib/fleet-collisions.sh` contains no file-overlap section, and `skills/bip-conductor/SKILL.md` invokes `bip epic collide` for that half. `docs/guides/shell-assumptions.md:11` and `skills/lib/clone-currency.sh:22` are updated to the new shape. `skills/lib/hazards/counting-a-failed-command.md` and `docs/guides/corpus-capture.md` keep their references — they record measured incidents against the script as it was.
3. The spans named in Phase 2 are replaced; the spans named in Out of scope are byte-identical.
4. `bip epic collide --help` matches the skill's invocation block verbatim.
5. `go test ./internal/gitx/...` passes; `go vet ./...` and `go fmt ./...` clean.

## Falsifier

If Phase 1 needs a judgement call per invocation, the prose returns as comments and nothing is saved — stop, and do not build Phase 2.

Checkable form, for Phase 2: falsified if any brief or verdict written after it lands states a consumer count without the command and commit it was taken at — concretely, if it quotes `26 consumers`, `47a35e7c`, `39 live consumers` or `ede049ec` from `skills/bip-epic/SKILL.md:368` instead of running `--symbol`. Quoting the reporting-vocabulary rules at `:370-387` is *intended* behaviour and is not a falsification.

## References

- `docs/guides/corpus-capture.md` — layer 3, and the correction that produced this scope.
- `#39` — proposes renaming `internal/flow`; the package siting above avoids adding to that cost.
- `#238` — the same move one layer over: replacing free-hand agent writes of `.epic-status.json` with a validated `bip epic` chokepoint. Coordinate on the `bip epic` subcommand namespace.
- `skills/lib/hazards/counting-a-failed-command.md` — the pipeline class the port must not reproduce.

matsen:
## Summary

After this PR `bip epic collide` performs the live-clone file-overlap check in Go with a scope it **derives** rather than defaults, `skills/lib/fleet-collisions.sh` keeps only its pane/process sections, and `skills/bip-epic/SKILL.md`'s symbol blast-radius block is an invocation.

The decisive reason is a live fail-open, not size or duplication. `fleet-collisions.sh` scoped itself with `ROOT="${1:-$HOME/re/pz}"` — optional, substituting on **empty** as well as unset, falling back to *another repository's* clone pool. Measured 2026-09-17 on `matsengrp/superfamily-pcp` and recorded in `skills/bip-conductor/SKILL.md`: an argument-less run reported clones `birch`/`cedar` and `COLLISION src/ml/stochastic_search.zig`, while an explicit-root re-run in the same minute reported `cobalt`/`copper`/`silver` and `COLLISION CLAUDE.md`. Every clone name, branch and file differed, nothing in the first output looks wrong, and it missed a real three-way `CLAUDE.md` collision that went on to produce two conflicting PRs. `bip-conductor/SKILL.md:529` recorded that as *"unfinished, not as fixed"*; this closes it.

The command takes **no scope argument**. It reads `clone_root` from the same `.epic-config.json` the rest of the fleet tooling reads, prints the root it resolved as its first line, and exits "could not check" if it cannot resolve one. An empty `--root` is an error, not a root.

## What changed

- **`bip epic collide`** — `cmd/bip/epic_collide.go` plus `internal/gitx/collide.go`, beside the existing `worktree.go` (*"small, tested wrappers over the git porcelain"*). Three sections: live branches and their touched files; files touched by more than one live clone; each live branch against what landed on `origin/main` since it forked.
- **`bip epic collide --symbol`** — `git grep` blast radius split into **code / prose / data** hits, stamped with the repo, commit and time, with the indirect-consumer hedge printed as its own last line.
- **Scope has two halves, and both are on the first output line.** The root comes from `clone_root`; the *universe* comes from `clone_names`, because a pool also holds clones nobody spawns into. Unlisted directories are reported `(unmanaged, not compared: ...)` — the same label `clone-currency.sh` uses — and declared-but-absent slots are listed too, so the report cannot silently examine fewer slots than it names.
- **Exit codes `0` clear / `10` found / `11` could not check.** `1` and `2` already mean general error and config error, and `cmd/bip/main.go` maps *every* `RunE` error to `1`, so a three-valued verdict cannot travel as an error: `RunE` returns `nil` and the command exits directly. `11` dominates `10` in `Report.Verdict` structurally — a caller acting on *found* believes the pool was fully examined.
- **`skills/lib/fleet-collisions.sh`**: 395 → 225 lines, 20,691 → 12,052 bytes. Sections 3/4/5 stay — they need a tmux socket, a process-subtree walk and `jq`. **Its own root is now REQUIRED and an empty one exits 2**, so the remaining sections no longer have the defect either.
- **Skill docs, same PR** (`CONSTITUTION.md` III; these skills are symlinked live, so a command-only landing leaves running sessions on a superseded procedure): the conductor's invocation block, the spawn-gate table row, `clone-currency.sh`'s sibling-shape warning, and `docs/guides/shell-assumptions.md`.

## Two deliberate divergences from the script, both toward checking more

- **A merge-base failure is uncheckable**, not a silent fallback to `origin/main`. The script fell back, reinstating the two-dot behaviour the merge base exists to avoid.
- **An unpushed branch is compared against the working tree**, not refused. A worker's branch is not pushed until the PR opens, so the script's *"no `origin/<branch>`, cannot compare"* made the **normal** state uncheckable: a smoke run on a two-clone pool exited 11 with two `UNCHECKABLE` lines and a real `COLLISION CLAUDE.md` in the same report.

  Both conductors were asked whether the fallback is necessary in their fleet, because if unpushed branches were rare it would be complexity to remove rather than keep. Both said keep it, with measurements. `phyz-conductor`: spawn `11:21:29Z` → PR opened `12:34:33Z`, a **~73-minute** unpushed window, measured from its own spawn log rather than from commit dates — a PR's `createdAt` can precede the earliest `committedDate` of its own commits, so commit dates cannot establish when work started. `sf-conductor`: it is unpushed **by construction** — its spawn template says *"Create branch ... from main"* at line 35 and *"Commit and push the branch"* only at the COMPLETION step at line 77 — measured **~90 minutes** for one slot today and most of a day for another.

  ⭐ **The sfpcp conductor role's framing is the decisive one and neither the issue nor I had it** (from an earlier `sf-conductor` session than the one currently holding that address): the check has no value *after* the PR opens.** By then the branch is pushed, visible, and about to be reviewed. Collisions matter precisely during the unpushed window, which is the one the refusal blanks out — so refusing there leaves the tool correct and useless. The report names which referent produced the number, because `mine=7` from a pushed branch and from a working tree are different claims.

## Where the issue's projection was wrong

`skills/bip-epic/SKILL.md` Step 4b went **15,117 → 14,804 bytes** — a 313-byte reduction, not the ~2,494 the issue forecast:

```
git show <ref>:skills/bip-epic/SKILL.md | awk '/^### Step 4b/{f=1} /^### Step 5/{f=0} f' | wc -c
```

The invocation is one line, but the measured instances around it had to stay: the command replaces the *counting*, not the *reading*, and two of the three recorded instances (an importing sibling that appends a duplicate key; a live branch that partitions the list by prefix) are invisible to any count. `docs/guides/corpus-capture.md` now records the corrected figure instead of the forecast. Across the three skill files the net is **−2,152 bytes** (script −8,639, epic −313, conductor **+6,800** for the verbatim `--help` paste and the universe rule below).

The `--help` paste is the one place this PR adds bytes on purpose, and `tests/integration/collide_help_test.go` fails if it drifts from the command in either direction — which is what makes `CONSTITUTION.md`'s *"skill behavior MUST match `--help` exactly"* mechanical for this command rather than a promise.

## What the fleet-scale run found, which the tests could not

⛔ **The command examined 21 directories where the pool has 17 slots, and two of the four extras were genuine clones that no worker is ever spawned into** — `nightly-ci` and `beagle-weekly-ci` on `matsengrp/phyz`, each named in a systemd unit and resetting hard to `origin/main` at run time, so a moving HEAD there is the *expected* state. The origin test correctly excluded a pinned `iqtree2` checkout and `_ignore` was not a clone; the CI clones share the pool's origin and were not excludable that way.

A conductor's action on a reported collision is to **delay or resequence a spawn**, so a false positive there costs a stalled slot for a reason nobody can reproduce. The run returned `clear` only because no clone had a live branch at that moment.

⭐ **The issue specified *"the same universe as the script"*, which was every directory — and `clone-currency.sh`, in the same directory, already scopes to `clone_names` and `bip-conductor/SKILL.md` already records why**: *"a check at full power over the wrong population."* I followed the spec instead of the sibling. Fixed in `e1be39e`.

➡ **This is why the fleet-scale run is not a formality after the tests pass.** Eleven mutations, real git fixtures, and every behaviour in the issue's list covered — and none of it had the wrong *population* in it, because I wrote the fixtures from the same spec that had the defect. `phyz-conductor` ran it against the real pool and read the first output line, which is what the first output line is for.

## Test plan

Real git fixtures — bare upstream, seed clone, pool root — reusing `initRepo` from `internal/gitx/worktree_test.go`. Every behaviour in the issue's carry-over list has a test that fails without it, and that was **verified by mutation rather than assumed**: eleven mutations were applied to `collide.go` and each was confirmed to turn the corresponding test red.

| mutation | test that catches it |
|---|---|
| diff `origin/main` instead of the merge base | `MergeBaseNotTwoDot` |
| leave git's rename detection on | `RenameOldPathAndSpaceInPath` |
| `-unormal` instead of `-uall` | `UntrackedFileCollides` |
| let found overwrite uncheckable | `UnreadableCloneIsUncheckableAndNotZero` |
| drop the foreign-clone exclusion | `ForeignCloneExcluded` |
| `[ -d "$d/.git" ]` instead of `rev-parse` | `MidRebaseWorktreeIsChecked` |
| cover only the `rebase-merge` backend | `MidRebaseWorktreeIsChecked` |
| report a degenerate ref as clear | `DegenerateRefIsUncheckable` |
| fall back to `origin/main` when merge-base fails | `NoMergeBaseIsUncheckableNotTwoDot` |
| make a failed fetch uncheckable instead of a warning | `FailedFetchWarnsAndContinues` |
| default an empty root to `$HOME/re/pz` | `EmptyRootIsNotARoot` |
| ignore `clone_names` and examine every directory | `ScopedToCloneNames` |

⚠ **One of those eleven initially passed, and that is the useful part of running them.** No test covered "a merge-base failure must be uncheckable rather than a fallback" — a behaviour named explicitly in the issue. `TestCollide_NoMergeBaseIsUncheckableNotTwoDot` exists because the mutation found the gap; reading the list and believing it was covered would not have.

Also covered: a one-clone pool does not report its own files as a collision; an empty pool root and one holding only `.spawn-prompts/` both say "nothing to check" and never print "clear"; an unreadable clone renders `files=UNREADABLE` with no `=0` on the line and still loses to nothing — uncheckable wins even when a real collision is also present; the pairwise section prints `live clones=N, pairs=M` so a bare "none" is never the whole answer; and `--root ""`, a missing config, an unparseable config and a config with no `clone_root` all fail to resolve rather than resolving to something else.

Flag combinations that would be silently ignored are refused: `--root` with `--symbol`, and `--path` without `--symbol`.

`go test ./...` passes; `go vet ./...` and `gofmt -l` clean.

## Two live-state corrections this PR also makes

Both were found by peers noticing that the mechanism they had been told about was not the one operating, and both are about artifacts symlinked into this worktree.

**`~/go/bin/bip` is a symlink to `/home/matsen/re/bipartite/bip`, an untracked build artifact — so there is no install step to lag.** `bip-conductor-spawn/SKILL.md` stated the opposite as a rule (*"a landed `bip` commit changes nothing until someone reinstalls"*), and that rule was measured and correct on 2026-09-14 — the installed binary really was 18 days old — before the wiring changed on 2026-09-19 03:47. The failure inverts under a symlink: the fleet's `bip` is whatever was last built in this worktree, from whatever branch was checked out; it changes with no commit, can go backwards, and **an unmerged branch's binary is live fleet-wide.** This PR demonstrated that — `bip epic collide` resolved for every session on the box because I rebuilt from this branch so the skill I had already made live would not prescribe a command that did not exist. The rule now opens with `ls -l "$(command -v bip)"` and describes both mechanisms.

**The evidence that rule was cited with is a wrong zero.** `bip-conductor/SKILL.md` used `strings ~/go/bin/bip | grep -cE '^--name$'` → `0` as proof the binary passed no `--name`. The flag reaches `claude` inside a format string, so it never appears as a standalone `strings` line and the anchored pattern returns 0 **whether or not the behaviour is present**. Same binary: `strings bip | grep -c "dangerously-skip-permissions --name"` → `1`, and `internal/flow/spawn/tmux.go:196` is `fmt.Sprintf("claude --dangerously-skip-permissions --name '%s'...")`. So the four bounced addresses it reports are real and now unexplained, and the install-lag story was fitted to the zero. Recorded that way rather than re-qualified — EVIDENCE-DISCIPLINE's own wrong-zero family, anchored-regex variant.

## Known limitation shipped on purpose, and named at the point of use

⛔ **#255: liveness comes from the branch name alone, so a clone sitting on a branch whose PR already MERGED counts as live.** Under squash-merge every commit on such a branch reads as unmerged, so the residual branch in a not-yet-reclaimed clone is reported forever — and the slot reclaim that would clear it is housekeeping a conductor does *after* running collide. It has already caught two conductors on `matsengrp/superfamily-pcp`.

It fires in both directions at once, and the second is what makes it survive a first look: a false `COLLISION` in section 2, and a false `OVERLAPS-LANDED` in section 3 whose printed remedy is *"rebase and take main's content for those files"* — for a branch that no longer exists for any purpose. A reader who correctly dismisses the collision still acts on the second, because it reads as a currency problem about a live branch rather than as a staleness artifact.

**Not fixed here.** The discriminator is `gh pr list --state merged --head <branch>`, which adds a network and auth failure mode to a command whose only network call is `git fetch origin main`, and the failure direction needs its own decision and tests — a `gh` call that fails must drive the pool to exit 11, never to a silently smaller live set. The limitation is in `--help`, so it travels with the binary, and in the skill with the check to run instead. **Merging this PR does not make that defect worse than it is on `main` today** — the port carried the name-only liveness test across from `fleet-collisions.sh` unchanged.

## DEFERRED

- **`git merge-tree --write-tree` per pair.** Out of scope per the issue — the script does no merge-conflict detection either, and `bip-epic/SKILL.md` records why it would not suffice alone (two PRs re-pinned different rows of one file, git merged them cleanly, and whether the citations resolved was decided by a separate check: *"a clean auto-merge is not a correct merge"*).
- **#256 — `bip-conductor-recover` resumes without `--name`.** Recovery converts the window-name address scheme into the `<cwd>-<suffix>` scheme without changing the window, so the one artifact a conductor composes an address from stays correct-looking while the address moves. Found by `sf-conductor` from a `/proc` read; corroborated here from `ListAgents` and against the recover skill's own invocation. This PR fixes the *documentation* of the two schemes, which was inverted; the behaviour is #256.
- **`fleet-collisions.sh` section 3 reports a deliberate conductor hold as `STALE`.** Raised by `phyz-conductor` against a live hold on `~/re/pz/cedar` (`stop_reason: conductor-hold-remote-jobs-live`, no worker behind it, remote jobs live on `orca03`). The predicate reads only `.issue` and `.updated_at` — no `stop_reason`, no `phase` — and the script's own guard states that *"the action on STALE is deletion"*. Real, and **not in this PR's scope**: it is a behaviour change to a section #252 explicitly keeps, it needs its own key decision (`stop_reason` prefix vs. an explicit boolean vs. a marker file) and its own tests. **Filed as #254**, which also records the decision it needs (which key marks a hold) rather than leaving it to the implementer.

Closes #252


matsen:
Closing in favour of #264. The decisive reason for this PR was the `${1:-$HOME/re/pz}` fail-open, and #264 closes it in two lines. The rest is about 1,300 lines of Go plus 1,250 of tests, and it adds 253 lines to `bip-conductor/SKILL.md`. That's more tooling than a file-overlap check needs, and it grows the corpus #262 is trying to shrink. #255, a false positive from squash-merged branches, shows the maintenance direction: each fix pulls in more state. The branch `252-epic-collide` is kept; reopen this PR if the Go port turns out to be needed.

